PR #55, opened 2014-10-03
Add a prompt to loopback generator for builtin models

raymondfeng:
The PR adds a prompt for loopback generator so that we can generate subclasses for built-in models.

It depends on https://github.com/strongloop/loopback-workspace/pull/153

/to @ritch @bajtos


superkhau:
@raymondfeng Another 2yr old PR, still relevant? Safe to close?

PR #74, opened 2014-12-02
Update create sample data script to be async

superkhau:
Connected to strongloop/loopback-workspace#170


superkhau:
Connected to #88

superkhau:
The proposed changes are included in #88 already. Closing.

Issue #80, opened 2015-01-08
Status middleware triggered for ALL URLs for default scaffolded app 


crandmck:
When you scaffold an app, by default the status middleware gets triggered not only for the root URL, but for _any_ URL, except for valid API URLs.  So `http://0.0.0.0:3000/`, `http://0.0.0.0:3000/foo`, `http://0.0.0.0:3000/foo/bar/`, and `http://0.0.0.0:3000/foo/bar.html` all trigger the status middleware.

This is a change to the behavior of the default scaffolded app, which used to trigger the status middleware ONLY on the root URL, `http://0.0.0.0:3000/`.   I'm not sure this is really a problem, but it does seem odd that it gets triggered for any invalid URL--you only see the error stack trace when you request an invalid URL that starts with `/api`.  I haven't found any way to actually generate a 404 error (UrlNotFoundError) in the default scaffolded app.

Perhaps this isn't an issue, since it's only with the default scaffolded app, but perhaps it would make more sense to make the status path be `/status` so it's not a "catchall" route.

@raymondfeng suggested I open an issue.

crandmck:
@bajtos PTAL when you have a moment.

bajtos:
@crandmck good catch, the fix is pending review.

crandmck:
Since https://github.com/strongloop/loopback/pull/996 was not merged, perhaps a compromise would be to bind the status middleware to a more specific route, for example `/status`, rather than the root, where it  becomes a "catchall".  Just a thought.

crandmck:
Oh wait, from https://github.com/strongloop/loopback-workspace/pull/182 it looks like you're going to put the status middleware back into `server/boot/root.js`... is that correct?

bajtos:
> Since strongloop/loopback#996 was not merged, perhaps a compromise would be to bind the status middleware to a more specific route, for example /status, rather than the root, where it becomes a "catchall". Just a thought.

This does not solve the problem, the middleware would become "catchall" for URLs starting with "/". Note that the main reason for this whole status URL is to provide a sensible response when opens `http://localhost:3000/` served by a new project. Before we decided to report status, the server was responding with 404.

> Oh wait, from strongloop/loopback-workspace#182 it looks like you're going to put the status middleware back into server/boot/root.js... is that correct?

Yes.

bajtos:
Fixed by https://github.com/strongloop/loopback-workspace/pull/182 and https://github.com/strongloop/generator-loopback/pull/82

[...]

PR #82, opened 2015-01-14
example: support both flavours of status route

bajtos:
The project template in loopback-workspace 3.6.0-3.6.5 incorrectly
registers `loopback.status()` middleware via `middleware.json`.
Versions until 3.5.x and after 3.6.6 register the status route
from a boot script.

This patch modifies the example generator to support both flavours.

See also https://github.com/strongloop/loopback-workspace/pull/182 and #80

/to @raymondfeng or @ritch please review

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -403,7 +403,14 @@ module.exports = yeoman.generators.Base.extend({
     var middleware = JSON.parse(this.readFileAsString(middlewareJSon));
 
     this._logPart('Remove status handler from "/"');
+    // loopback-workspace 3.6.0-3.6.5
     delete middleware.routes['loopback#status'];
+
+    // loopback-workspace <3.6.0 and 3.6.6+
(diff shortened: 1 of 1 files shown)


bajtos:
@ritch @raymondfeng ping, PTAL

ritch:
So this removing the matter where it is and expects the workspace to add it?

bajtos:
> So this removing the matter where it is and expects the workspace to add it?

I don't understand the question.

The example generator needs to remove the default `/` route handler and replace it with a static middleware serving client assets.

This patch modified the generator to support both flavours of the default `/` route handler registration, both flavours of the loopback-workspace template - the old one and the new one.

bajtos:
@raymondfeng @ritch ping, any more comments? may I merge this patch?

ritch:
LGTM

bajtos:
Released as `v1.7.1`.

PR #84, opened 2015-02-06
Rename `gitignore` to `.gitignore`

superkhau:
This fix renames the `gitignore` file back to `.gitignore`. This is required
because NPM renames `.gitignore` to `.npmignore` when publishing.


bajtos:
See my comment in strongloop/loopback-workspace#186, the problem of `.gitignore` should be fixed by loopback-workspace in a way that does not require any changes in the consumers of the API.

Issue #86, opened 2015-02-17
Use native methods to copy .gitignore


superkhau:
Refactor `injectWorkspaceCopyRecursive` to use native methods to copy `gitignore` to `.gitignore`. See https://github.com/strongloop/loopback-workspace/pull/186

superkhau:
Closing, duplicate in PR https://github.com/strongloop/generator-loopback/pull/87

PR #87, opened 2015-02-18
Refactor `injectWorkspaceCopyRecursive`

superkhau:
Use native `this.copy` when source file is not a directory. This allows us to
copy `gitignore` to `.gitignore` when scaffolding new projects.

### DIFF
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -17,6 +17,7 @@ describe('loopback:app generator', function() {
   // we don't have to test it again.
 
   var EXPECTED_PROJECT_FILES = [
+    '.gitignore',
     'package.json',
 
     'server/config.json',
--- a/app/index.js
+++ b/app/index.js
@@ -5,6 +5,7 @@ var chalk = require('chalk');
 var workspace = require('loopback-workspace');
 var Workspace = workspace.models.Workspace;
 
+var fs = require('fs');
 var path = require('path');
 
 var actions = require('../lib/actions');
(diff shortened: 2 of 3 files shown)


superkhau:
See https://github.com/strongloop/generator-loopback/issues/86

superkhau:
@bajtos Review please.

superkhau:
@bajtos The CI test will fail until you merge the changes in https://github.com/strongloop/loopback-workspace/pull/186 as this test depends on the `gitignore` existing in `loopback-workspace` (since `this.copy` has nothing to copy until then).

test/app.test.js
```diff
@@ -17,6 +17,8 @@ describe('loopback:app generator', function() {
   // we don't have to test it again.
 
   var EXPECTED_PROJECT_FILES = [
+    '.gitignore',
+
```
bajtos: The entries are grouped by the directory, please remove the empty L21.

bajtos:
One minor comment, the rest LGTM.

> The CI test will fail until you merge the changes in strongloop/loopback-workspace#186 as this test depends on the gitignore existing in loopback-workspace (since this.copy has nothing to copy until then).

Note that our CI system does not need the dependent PR to be landed. When you use the same branch name in both projects, CI will install the right version of the dependent including the pending changes. You can see that the "default" merge build is green.

superkhau:
@bajtos Any reason as to why the `continuous-integration/travis-ci/pr` and `continuous-integration/travis-ci/push` fails then?

superkhau:
@slnode test please

superkhau:
After discussion with @bajtos on Slack, it was concluded that Travis builds using `npm install`, and since the `loopback-workspace` hasn't been released yet, CI build will fail until then.

superkhau:
@slnode test please

superkhau:
Connected to strongloop/strong-arc/issues/685

PR #88, opened 2015-02-20
Use async features in example generator

superkhau:
- Change `app.importing` to `app.booting`
- Use callbacks instead of emitting `import done`
- Replace `import done` event with the `ready` event

Connected to strongloop/generator-loopback#74

### DIFF
--- a/example/templates/server/sample-data/import.js
+++ b/example/templates/server/sample-data/import.js
@@ -53,10 +53,6 @@ module.exports = function(app, cb) {
   });
 };
 
-if (require.main === module) {
-  // Run the import (server runs it automatically during boot)
-  var app = require('../server');
-  app.on('import done', function(err) {
-    process.exit(err ? 1 : 0);
-  });
-}
+if (require.main === module)
+  // The import runs automatically during the boot process.
+  require('../server');
--- a/example/templates/server/test/rest-api.test.js
+++ b/example/templates/server/test/rest-api.test.js
@@ -8,11 +8,10 @@ var assert = require('assert');
 
 before(function importSampleData(done) {
   this.timeout(50000);
-  if (app.importing) {
-    app.on('import done', done);
-  } else {
+  if (app.booting)
+    app.on('booted', done);
+  else
     done();
-  }
 });
 
 function json(verb, url) {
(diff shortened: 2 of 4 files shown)


superkhau:
@bajtos Review please.

example/templates/server/sample-data/import.js
```diff
@@ -53,10 +53,6 @@ module.exports = function(app, cb) {
...
-if (require.main === module) {
-  // Run the import (server runs it automatically during boot)
-  var app = require('../server');
-  app.on('import done', function(err) {
-    process.exit(err ? 1 : 0);
-  });
-}
+if (require.main === module)
+  // The import runs automatically during the boot process.
+  require('../server');
```
bajtos: Will this exit the process with a non-zero status code when the import fails?

example/templates/server/test/rest-api.test.js
```diff
@@ -8,11 +8,10 @@ var assert = require('assert');
 
 before(function importSampleData(done) {
   this.timeout(50000);
-  if (app.importing) {
-    app.on('import done', done);
-  } else {
+  if (app.booting)
+    app.on('ready', done);
```
bajtos: You should use `booted` event here.

example/templates/server/sample-data/import.js
superkhau: Yes, as long as `err` is returned in the callback (we `throw` by default in https://github.com/strongloop/loopback-workspace/commit/223758f6cb4051e4791c52ce6f8f2f88a4200f2d#diff-383e9ce062435fb34a4326cdf175c907R16 causing an `unhandledException` event causing Node to return 1 for the exit code). I did a smoke test by changing the name of the JSON file in `import.js` deliberately causing a missing file type error then checking the exit status in bash via `echo $?` which returned 1.

superkhau:
@bajtos Review please.

bajtos:
The code LGTM. Please update the commit message before landing, "Replace `import done` event with the `ready` event" is no longer true.

superkhau:
TY, commit message removed.

bajtos:
Cool, please land the patch yourself. Do you have permissions to publish a new version to npmjs.org? Does this depend on a new version of loopback-workspace (see https://github.com/strongloop/loopback-workspace/pull/172)? In that case we should bump up the minimum required version in package.json too.

superkhau:
TY, yes I have permission. Will bump workspace version and merge.

PR #91, opened 2015-02-24
Upgrade strong-cached-install to ^2.0 for io.js

bajtos:


bajtos:
@rmg @ritch There is one thing I am not sure about. Should we release this as a major version, because we are dropping compatibility with Node v0.8? On the other hand, AFAIK, we never explicitly supported Node v0.8, in which case this can be considered a patch version. Thoughts?

bajtos:
Actually, this is probably not a problem, because generator-loopback replaces the method calling `ncp` with a custom implementation calling yeoman's copy.

Issue #90, opened 2015-02-24
Add generator for boot scripts


superkhau:
I would like to add a generator to scaffold boot scripts, something like:

```
slc loopback:boot
What kind of boot script would you like to generate?
>o sync
o async
What do you want to name your boot script? <enter name>
[generate <name>.js] in /server/boot with the following contents:
#sync
module.exports = function(app) {
  
}
#async
module.exports = function(app, cb) {
  
  cb();
};
```

@bajtos @raymondfeng @ritch Thoughts?

Connected to #93

bajtos:
Sure, why not. Perhaps `loopback:boot-script` would be a better name.

At the moment, loopback-workspace does not provide any API for boot scripts. I don't think it makes much sense to introduce such API, thus I am proposing to keep the implementation of this feature in the generator only.

raymondfeng:
+1. We should also add 'loopback:method' to generate custom remote method, which is even more useful.

superkhau:
@raymondfeng +1 Remote method is probably even more important. We should probably go with `loopback:remote-method` (sticking with @bajtos `loopback:boot-script` suggestion from above).

We should probably also do `loopback:operation-hook` to make it feature complete.

superkhau:
>+1. We should also add 'loopback:method' to generate custom remote method, which is even more useful.

Track at https://github.com/strongloop/generator-loopback/issues/94

>We should probably also do loopback:operation-hook to make it feature complete.

Track at https://github.com/strongloop/generator-loopback/issues/95

superkhau:
PR at https://github.com/strongloop/generator-loopback/pull/93

superkhau:
@bajtos Removed `yo-rc.json` related code. PTAL again.

PR #93, opened 2015-03-05
Add boot script generator

superkhau:
- Add code for `slc loopback:boot script` command
- Add unit tests

Connected to #90 

- [x] Code
- [x] Code Review
- [x] Docs (https://github.com/strongloop/generator-loopback/wiki/Boot-script-generator)
- [x] Examples (https://github.com/strongloop/loopback-example-boot-scripts)

### DIFF
--- a/boot-script/index.js
+++ b/boot-script/index.js
@@ -1,0 +1,65 @@
+'use strict';
+var helpers = require('../lib/helpers');
+var path = require('path');
+var yeoman = require('yeoman-generator');
+
+module.exports = yeoman.generators.Base.extend({
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+
+    this.argument('name', {
+      desc: 'Name of the boot script to create.',
+      required: false,
+      optional: true,
+      type: String
+    });
+  },
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  askForName: function() {
+    var done = this.async();
+
+    if (this.name) return done();
+
+    var question = {
+      name: 'name',
+      message: 'Enter the script name (without `.js`):',
+      default: this.name,
+      validate: helpers.validateName
+    };
+
+    this.prompt(question, function(answer) {
+      this.name = answer.name;
+      done();
+    }.bind(this));
+  },
+
+  askForType: function() {
+    var done = this.async();
+
+    var question = {
+      name: 'type',
+      message: 'What type of boot script do you want to generate?',
+      type: 'list',
+      choices: ['async', 'sync'],
+      default: 'async'
+    };
+
+    this.prompt(question, function(answer) {
+      this.type = answer.type;
+      done();
+    }.bind(this));
+  },
+
+  generate: function() {
+    var source = this.templatePath(this.type + '.js');
+
+    var targetPath = path.normalize('/server/boot/' + this.name + '.js');
+    var target = this.destinationPath(targetPath);
+
+    this.copy(source, target);
+  }
+});
--- a/boot-script/templates/async.js
+++ b/boot-script/templates/async.js
@@ -1,0 +1,10 @@
+module.exports = function(app, cb) {
+  /*
+   * The `app` object provides access to a variety of LoopBack resources such as
+   * models (e.g. `app.models.YourModelName`) or data sources (e.g.
+   * `app.datasources.YourDataSource`). See
+   * http://docs.strongloop.com/display/public/LB/Working+with+LoopBack+objects
+   * for more info.
+   */
+  process.nextTick(cb); // Remove if you pass `cb` to an async function yourself
+};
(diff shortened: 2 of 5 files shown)


superkhau:
@bajtos Code review please.

/cc @raymondfeng @ritch

boot-script/index.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+
+  askForName: function() {
+    var done = this.async();
+
+    if (this.name) return done();
+
+    var question = {
+      name: 'name',
+      message: 'Enter the model name:',
```
raymondfeng: Please fix the message

boot-script/index.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+
+  askForName: function() {
+    var done = this.async();
+
+    if (this.name) return done();
+
+    var question = {
+      name: 'name',
+      message: 'Enter the script name (without `.js`):',
```
superkhau: @raymondfeng Fixed.

boot-script/templates/async.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(app, cb) {
+  
+
```
bajtos: Two empty lines, please keep only one of them.

app/index.js
```diff
@@ -154,6 +154,11 @@ module.exports = yeoman.generators.Base.extend({
     this.directory('.', '.');
   },
 
+  generateYoRc: function() {
+    this.log('Generating .yo-rc.json');
+    this.config.save();
+  },
+
```
bajtos: What do you need `.yo-rc.json` for? I don't see any other usage of `this.config` in this patch.

bajtos: Can we revert these 5 lines?

bajtos:
In the future, when we implement support for multiple facets (e.g. configuring loopback app in "client"), we'll need to support "/client/boot" too. 

Right now, I just want to check that the current design won't make it difficult to support that?

app/index.js
superkhau: .yo-rc.json allows the boot script command (any yo command (AFAIK) from any subdir in the project. This means we can run `yo loopback:boot-script` without actually being in the project root (because yeoman looks for `.yo-rc.json` to determine the project root). IMO this should've been there a long time ago. Opinions?

boot-script/templates/async.js
superkhau: Fixed.

superkhau:
Yes, a simple refactor to accept a different destination dir will be possible. In that future case you describe, I'm thinking we can ask one more question (Is this a client or server side boot script? (show a list of two items client/server).

superkhau:
@bajtos Added your review changes, PTAL again.

app/index.js
bajtos: I see. Can you submit this change in a new commit please and include a unit-test for that? (A new pull request is ok too, so that this PR does not have to wait for that.)

superkhau: Sure, I will pull this feature in a separate PR. I have no idea how to unit test this, but we can discuss it over there instead. ;)

boot-script/templates/async.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function(app, cb) {
+  
+  cb();
```
raymondfeng: Maybe better to have the following:
```js
/* 
 * The `app` object allows you to access a model by name (e.g. MyModel) as `app.models.MyModel`
 * or a data source by name (e.g. MyMongoDB) as `app.dataSources.MyDataSource`
 */
process.nextTick(cb); // Please remove it if you already pass `cb` to an async function
```

superkhau: Good idea. Updating, except for the comment which is too long to fit in 80 chars. I also add the app object comments to `sync.js`.

boot-script/templates/async.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function(app, cb) {
+  /*
+   * The `app` object provides access to a variety of LoopBack resources such as
+   * models (e.g. `app.models.YourModelName`) or data sources (e.g.
+   * `app.datasources.YourDataSource`). See
+   * http://docs.strongloop.com/display/public/LB/Working+with+LoopBack+objects
+   * for more info.
+   */
+  process.nextTick(cb); // Remove if you pass `cb` to an async function yourself
```
superkhau: @raymondfeng PTAL

boot-script/templates/sync.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function(app) {
+  /*
+   * The `app` object provides access to a variety of LoopBack resources such as
+   * models (e.g. `app.models.YourModelName`) or data sources (e.g.
+   * `app.datasources.YourDataSource`). See
+   * http://docs.strongloop.com/display/public/LB/Working+with+LoopBack+objects
+   * for more info.
+   */
```
superkhau: @raymondfeng PTAL

raymondfeng: Link is nice.

bajtos:
LGTM, please wait for @raymondfeng to approve too.

raymondfeng:
LGTM

superkhau:
@crandmck Merged.

PR #96, opened 2015-03-10
Create .yo-rc.json when scaffolding new apps

superkhau:
Yeoman looks for `.yo-rc.json` (out-of-box) to determine the project root when
executing scripts. This means generators and sub-generator can be run in child
directories (not only from the project root).

Connected to strongloop-internal/scrum-loopback#207

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -154,6 +154,11 @@ module.exports = yeoman.generators.Base.extend({
     this.directory('.', '.');
   },
 
+  generateYoRc: function() {
+    this.log('Generating .yo-rc.json');
+    this.config.save();
+  },
+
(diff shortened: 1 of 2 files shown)


superkhau:
@bajtos PTAL

bajtos:
It would be nicer to test the desired behaviour ("slc loopback:model" can be run in a subdirectory) rather than the implementation details (".yo-rc.json" is created), but I can live with this too.

LGTM.

superkhau:
Gonna merge this first to make sure our users get this feature before writing that extra unit test. I will need your help in figuring out how to write that unit test to simulate `slc loopback:model`. Let's continue this conversation at #97

PR #100, opened 2015-04-29
Change slc run to slc start in end next steps

duffn:
Change _Next Steps_ instructions to recommend `slc start` instead of `slc run` in alignment with documentation.

http://docs.strongloop.com/display/NODE/slc+run
http://docs.strongloop.com/display/public/LB/Running+apps+with+slc

Connect to #101

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -182,10 +182,7 @@ module.exports = yeoman.generators.Base.extend({
     this.log(chalk.green('    $ slc strongops'));
     this.log();
     this.log('  Run the app');
-    if (cmd === 'yo')
-      this.log(chalk.green('    $ node .'));
-    else
-      this.log(chalk.green('    $ ' + cmd + ' start'));
+    this.log(chalk.green('    $ node .'));
     this.log();
   }
 });
--- a/app/index.js
+++ b/app/index.js
@@ -178,8 +178,8 @@ module.exports = yeoman.generators.Base.extend({
     this.log('  Create a model in your app');
     this.log(chalk.green('    $ ' + cmd + ' loopback:model'));
     this.log();
-    this.log('  Optional: Enable StrongOps monitoring');
-    this.log(chalk.green('    $ slc strongops'));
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
@sam-github you are more familiar with `slc run` and `slc start` than me. What's your opinion on the change presented here?

bajtos:
@slnode ok to test

bajtos:
> Change Next Steps instructions reported by `slc loopback` to recommend `slc start` instead of `slc run` in alignment with documentation.

@sam-github @altsang @ijroth do you have any objections against landing this change?

slnode:

altsang:
think we need to clarify the steps...

1. if you are in dev and you want to simply run the app  - `node server.js`
2. if you are in dev and you want to test the application under the StrongLoop Process Manager `slc start` in the app directory

the documentation could list both, but as the default next steps for `slc loopback` i would print out message 1 only

sam-github:
+1

slnode:

crandmck:
I agree with what @altsang said.

> the documentation could list both, but as the default next steps for slc loopback i would print out message 1 only

Since this would presumably be early in the dev process, then `node ` is more appropriate than `slc start`, which doesn't give any feedback to the console (e.g. is the app running, stack traces, err messages, etc).  So if we have to choose one, then I suggest we say use `node`.

**However** using `node server.js` does not work, because the main script is in the `server` directory.  You have to do either `node server/server.js` or `node .`.  The docs use the latter, so I suggest we go with that.

crandmck:
Commented in #100.  Really, using `node` makes more sense in this context IMHO.

altsang:
@crandmck yes - good catch `node server/server.js` or `node .`

bajtos:
Landed via 8605c16, I have fixed the code to fit all lines into 80 chars, squashed the commits and improved the commit message.

Thank you @duffn for the contribution.

PR #104, opened 2015-06-08
Add a generator for middleware

raymondfeng:
/to @bajtos 

Depends on https://github.com/strongloop/loopback-workspace/pull/203

### DIFF
--- a/middleware/index.js
+++ b/middleware/index.js
@@ -1,0 +1,236 @@
+'use strict';
+var chalk = require('chalk');
+var yeoman = require('yeoman-generator');
+
+var wsModels = require('loopback-workspace').models;
+
+var actions = require('../lib/actions');
+var helpers = require('../lib/helpers');
+var validateName = helpers.validateName;
+
+function toNumberedList(items) {
+  var lastIndex = items.length - 1;
+  return items.map(function(item, index) {
+    var name = item;
+    if (index !== lastIndex) {
+      // Last item should not be numbered
+      name = (index + 1) + '. ' + item;
+    }
+    return {
+      name: name,
+      value: item
+    };
+  });
+}
+
+var OTHER_PHASE = '(custom phase)';
+var LAST_PHASE = '(last phase)';
+
+module.exports = yeoman.generators.Base.extend({
+  // This generator does not track file changes via yeoman,
+  // as loopback-workspace is editing (modifying) files when
+  // saving project changes.
+
+  loadProject: actions.loadProject,
+
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+
+    this.argument('name', {
+      desc: 'Name of the middleware to create.',
+      required: false,
+      type: String
+    });
+  },
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  loadPhases: function() {
+    var done = this.async();
+    wsModels.Middleware.getPhases(function(err, list) {
+      if (err) {
+        return done(err);
+      }
+
+      this.phases = list.map(function(m) {
+        return m.phase;
+      });
+
+      done();
+    }.bind(this));
+  },
+
+  askForName: function() {
+    var done = this.async();
+
+    var prompts = [
+      {
+        name: 'name',
+        message: 'Enter the middleware name:',
+        default: this.name,
+        validate: function(name) {
+          return name && validateName(name);
+        }
+      }
+    ];
+
+    this.prompt(prompts, function(props) {
+      this.name = props.name;
+      done();
+    }.bind(this));
+
+  },
+
+  askForPhase: function() {
+    var done = this.async();
+
+    var displayName = chalk.yellow(this.name);
+
+    var prompts = [
+      {
+        name: 'phase',
+        message: 'Select the phase for ' + displayName + ':',
+        type: 'list',
+        default: 'routes',
+        choices: toNumberedList(this.phases.concat([OTHER_PHASE]))
+      },
+      {
+        name: 'customPhase',
+        message: 'Enter the phase name:',
+        validate: validateName,
+        when: function(answers) {
+          return answers.phase === OTHER_PHASE;
+        }
+      },
+      {
+        name: 'nextPhase',
+        message: 'Select the phase before which the new one will be inserted:',
+        type: 'list',
+        default: 'routes',
+        choices: toNumberedList(this.phases.concat([LAST_PHASE])),
+        when: function(answers) {
+          return answers.phase === OTHER_PHASE;
+        }
+      },
+      {
+        name: 'subPhase',
+        message: 'Select the sub phase for ' + displayName + ':',
+        type: 'list',
+        default: '',
+        choices: [
+          {name: '1. before', value: 'before'},
+          {name: '2. regular', value: ''},
+          {name: '3. after', value: 'after'}
+        ]
+      },
+    ];
+
+    this.prompt(prompts, function(props) {
+      this.phase = props.customPhase || props.phase;
+      this.customPhase = props.customPhase;
+      this.nextPhase = props.nextPhase;
+      this.subPhase = props.subPhase;
+      done();
+    }.bind(this));
+  },
+
+  promptForPaths: function() {
+    var displayName = chalk.yellow(this.name);
+    this.log('Specify paths for ' + displayName + ':');
+  },
+
+  askForPaths: function() {
+    var done = this.async();
+    this.log('Enter an empty path name when done.');
+    var prompts = [
+      {
+        name: 'path',
+        message: 'Path uri:',
+        validate: function(input) {
+          if (input) {
+            if (input.indexOf('/') !== 0) {
+              return 'Path must start with /';
+            } else {
+              return true;
+            }
+          } else {
+            return true;
+          }
+        }
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      if (answers.path == null || answers.path === '') {
+        return done();
+      }
+
+      this.paths = this.paths || [];
+      this.paths.push(answers.path);
+      this.log('Let\'s add another path.');
+      this.askForPaths();
+    }.bind(this));
+  },
+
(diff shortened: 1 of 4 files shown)


bajtos:
@raymondfeng as we discussed in Hangout, I have to high-level concerns:
 1. When adding a new (custom) middleware phase, user needs to specify the place (index, slot) where to insert the new phase. Adding the new phase at the end is almost certainly not what the user wants, because the `final` phase installs catch-all handler converting all requests to 404.
 2. In order to achieve the above, we will need to specify `middleware.order`. This may be tedious to use from the client, it may be better to change workspace API to better support this use case.

app/index.js
```diff
@@ -16,6 +16,10 @@ var pkg = require('../package.json');
 module.exports = yeoman.generators.Base.extend({
   constructor: function() {
     yeoman.generators.Base.apply(this, arguments);
+    // Skip file conflict detection
+    if (this.conflicter) {
+      this.conflicter.force = true;
+    }
```
bajtos: What is the purpose of this change? I went quite far in the implementation to actually enable conflict detection, as I find it a useful feature. Could you please open a new PR to make & discuss this change?

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+function listChoices(items) {
+  var choices = [];
+  items.forEach(function(item, index) {
+    choices.push({
+      name: (index + 1) + '. ' + item,
+      value: item
+    });
+  });
+  return choices;
```
bajtos: Please simplify and use `items.map()` instead of `items.forEach()` + `choices.push()`.

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+}
+
+module.exports = yeoman.generators.Base.extend({
+  // NOTE(bajtos)
```
bajtos: Please remove my name from the note as I am not the author of the note in this particular file. I think we can remove this line entirely and keep just the comment.

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+
+      var phases = {};
+      list.forEach(function(c) {
+        if (c.phase) {
+          phases[c.phase] = c.phase;
+        }
+      });
+
+      this.phases = Object.keys(phases);
```
bajtos: IIUC, this assumes that `wsModels.Middleware.find` returns the middleware entries sorted by the phase index in the middleware file. Is that a correct assumption to make?

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+    var displayName = chalk.yellow(this.name);
+
+    var prompts = [
+      {
+        name: 'phase',
+        message: 'Select the phase for ' + displayName + ':',
+        type: 'list',
+        default: 'routes',
+        choices: listChoices(this.phases.concat(['__other__']))
```
bajtos: I would personally use a more descriptive name than `listChoices()`, e.g. `toNumberedList()`.

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+        name: 'customPhase',
+        message: 'Enter the phase name:',
+        validate: validateName,
+        when: function(answers) {
+          return answers.phase === '__other__';
+        }
+      },
+      {
+        name: 'nextPhase',
+        message: 'Select the phase right after the custom phase:',
```
bajtos: > Select the phase right after the custom phase

I am not sure how easy it will be to understand the purpose of this prompt. @ritch @crandmck could you please comment?

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+    }.bind(this));
+  },
+
+  middleware: function() {
+    var done = this.async();
+    var config = {
+      name: this.name,
+      phase: this.phase,
+      paths: this.paths,
+      params: this.params || {},
```
bajtos: I think `false` is a valid `params` value. You may want to use a different check here.

bajtos: Also note that `params` is an array of arguments to pass to the factory function. The case where `params` is an object is a special syntax sugar. It may be worth to use `[]` as default instead of `{}`.

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+        } else {
+          process.nextTick(done);
+        }
+      },
+      function addMiddleware(done) {
+        wsModels.Middleware.create(config, function(err, inst) {
+          helpers.reportValidationError(err, self.log);
+          if (!err && inst) {
+            self.log(
+              'Middleware %s is added to phase %s.', inst.name, inst.phase);
```
bajtos: Middleware *was* added?

test/middleware.test.js
```diff
@@ -0,0 +1,70 @@
...
+    var path = '../../middleware';
+    var name = 'loopback:middleware';
+    var gen = common.createGenerator(name, path, [], dsArgs, {});
+    return gen;
+  }
+
+  function readMiddlewaresJsonSync(facet) {
+    var filepath = path.resolve(SANDBOX, facet || 'server', 'middleware.json');
+    var content = fs.readFileSync(filepath, 'utf-8');
+    return JSON.parse(content);
```
bajtos: You can use `fs.readJsonFileSync(filepath)` here, providing that you change `var fs` to `require('fs-extra')`.

middleware/index.js
bajtos: I find the display name `__other__` rather ugly, can you please provide a nicer text to display, e.g. `(create a new phase)`? 

IIRC:

```js
this.phases.concat({ name: '(create a new phase)', value: '__other__' });
```

middleware/index.js
```diff
@@ -0,0 +1,227 @@
...
+        when: function(answers) {
+          return answers.phase === '__other__';
+        }
+      },
+      {
+        name: 'nextPhase',
+        message: 'Select the phase right after the custom phase:',
+        type: 'list',
+        default: 'routes',
+        choices: listChoices(this.phases.concat(['__last__'])),
```
bajtos: Ditto as in https://github.com/strongloop/generator-loopback/pull/104/files#r32245016, please use a nicer display text.

bajtos:
@raymondfeng the new generator looks mostly good, see my comments above.

There is one more thing to consider & fix. Each phase (e.g. "routes") has three sub-phases ("routes:before", "routes", "routes:after"). While most of the before/after sub-phases are not listed in the `middleware.json` file by default, I think the generator should still support them all.

It gets more tricky when the user wants to add a new phase (e.g. "audit") and add a middleware to a sub-phase of this new phase (e.g. "audit:after"). Also when selecting a position for this new phase, it is not valid to insert a phase between sub-phases, e.g. "auth" -> "audit" -> "auth:after".

I am proposing the following solution:

 1. The list of phases should merge all sub-phases into a single item: "auth:before", "auth" and "auth:after" are merged into a single item "auth".
 2. When selecting a phase (either where to add the middleware or where to add the new phase), list only these merged items: "initial", "auth", etc.
 3. Once the user has selected the phase, ask them for a sub-phase (regular, before, after).

bajtos:
Last but not least, please make the CI build green before landing this patch.

middleware/index.js
crandmck: It's hard for me to say without seeing it in context....  When will the user see this prompt?

middleware/index.js
raymondfeng: Fixed

middleware/index.js
raymondfeng: Fixed

middleware/index.js
bajtos: @crandmck 
> It's hard for me to say without seeing it in context.... When will the user see this prompt?

When the user decides to use add the new middleware to a new phase that does not exist yet (option `__other__` in the screenshot below), then the generator asks for a new phase name and then asks for a place where to insert this new phase.

```
? Enter the middleware name: foobar
? Select the phase for foobar: 10. __other__
? Enter the phase name: audit
? Select the phase right after the custom phase:
  7. parse
  8. routes
  9. session
❯ 10. __last__
  1. auth
  2. files
  3. final
```

middleware/index.js
bajtos: FWIW, I find it slightly misleading that this "other" option has a sequence number too. Since this number is the last one, it looks like the new phase will come after "final".

I am proposing to remove the sequence number from the "other" option, i.e.

```
  7. parse
  8. routes
  9. session
> (create a new phase)
```

bajtos:
@raymondfeng I have discovered another problem while playing with the generator (revision 908f5ac). It seems that middleware phases are printed in incorrect order.

The initial viewport:

```
? Enter the middleware name: foobar
? Select the phase for foobar: (Use arrow keys)
  5. initial
  6. initial:before
  7. parse
❯ 8. routes
  9. session
  10. __other__
  1. auth
```

The rest:

```
? Select the phase for foobar:
  1. auth
  2. files
  3. final
❯ 4. final:after
  5. initial
  6. initial:before
  7. parse
```

Here is my server/middleware.json
```
{
  "initial:before": {
    "loopback#favicon": {}
  },
  "initial": {
    "compression": {}
  },
  "session": {
  },
  "auth": {
  },
  "parse": {
  },
  "routes": {
  },
  "files": {
  },
  "final": {
    "loopback#urlNotFound": {}
  },
  "final:after": {
    "errorhandler": {}
  }
}
```

bajtos:
FWIW, I find it slightly misleading that the "[custom phase]" option  has a sequence number too. Since this number is the highest one printed, it looks like the new phase will come after "final".

I am proposing to remove the sequence number from the "[custom phase]" option, i.e.

```
  7. parse
  8. routes
> 9. session
  [custom phase]
```

raymondfeng:
@bajtos Comments addressed. PTAL.

bajtos:
@raymondfeng looks mostly good. could you please fix the failing CI build?

crandmck:
Thanks @raymondfeng for adding docs! http://docs.strongloop.com/display/LB/Middleware+generator

PR #110, opened 2015-08-14
Upgrade deps to yeoman-generator 0.20.x

raymondfeng:
/to @bajtos 
/cc @rmg 

The main effort is to upgrade to yeoman-generator 0.20.x which has quite a few breaking changes. This PR also fixes the build failure in master.

### DIFF
--- a/boot-script/index.js
+++ b/boot-script/index.js
@@ -57,7 +57,8 @@ module.exports = yeoman.generators.Base.extend({
   generate: function() {
     var source = this.templatePath(this.type + '.js');
 
-    var targetPath = path.normalize('/server/boot/' + this.name + '.js');
+    // yeoman-generator 0.20.x doesn't like the leading /
+    var targetPath = path.normalize('server/boot/' + this.name + '.js');
     var target = this.destinationPath(targetPath);
 
     this.copy(source, target);
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -115,12 +115,12 @@ actions.loadModels = function() {
  * the actual installation is invoked later as part of the "installing" phase.
  */
 actions.installDeps = function() {
-  this._skipInstall = this.options['skip-install'];
+  // yeoman-generator 0.20.x checks this.options.skipInstall
+  this.options.skipInstall = this.options['skip-install'];
 
   this.installDependencies({
     npm: true,
-    bower: false,
-    skipInstall: this._skipInstall
+    bower: false
   });
 };
 
--- a/package.json
+++ b/package.json
@@ -27,21 +27,21 @@
     "strongloop"
   ],
   "dependencies": {
-    "async": "^0.9.0",
-    "chalk": "^0.5.0",
-    "debug": "^2.1.0",
-    "inflection": "^1.3.8",
+    "async": "^1.4.2",
+    "chalk": "^1.1.0",
+    "debug": "^2.2.0",
+    "inflection": "^1.7.1",
     "loopback-swagger": "^2.0.0",
-    "loopback-workspace": "^3.15.0",
-    "request": "^2.47.0",
-    "yeoman-generator": "^0.18.7",
-    "yosay": "^1.0.0"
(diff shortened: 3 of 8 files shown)


lib/actions.js
```diff
@@ -115,12 +115,12 @@ actions.loadModels = function() {
...
 actions.installDeps = function() {
-  this._skipInstall = this.options['skip-install'];
+  // yeoman-generator 0.20.x checks this.options.skipInstall
+  this.options.skipInstall = this.options['skip-install'];
 
   this.installDependencies({
     npm: true,
-    bower: false,
-    skipInstall: this._skipInstall
+    bower: false
```
rmg: Does this need to run in browsers? If not, leaving the trailing `,` reduces noise in the commit history :-)

test/common.js
```diff
@@ -29,7 +29,11 @@ exports.createDummyProject = function(dir, name, done) {
 
 exports.resetWorkspace = function(done) {
   async.each(workspace.models(), function(model, cb) {
-    model.destroyAll(cb);
+    if (typeof model.destroyAll === 'function') {
```
rmg: @ritch why don't all workspace models have a `.destroyAll()` method in loopback-workspace@3.16.x?

package.json
```diff
@@ -27,21 +27,21 @@
...
   },
   "devDependencies": {
-    "chai": "^1.10.0",
-    "fs-extra": "^0.12.0",
-    "jshint": "^2.5.8",
-    "mocha": "^2.0.1",
+    "chai": "^3.2.0",
+    "fs-extra": "^0.23.1",
+    "jshint": "^2.8.0",
+    "mocha": "^2.2.5",
```
rmg: The following dependency upgrades appear to be unnecessary because they won't actually change the version that gets installed:
- debug
- inflection
- loopback-workspace
- request
- yosay
- jshint
- mocha

Unless this PR includes changes that actually requires features added in the new minimum versions, all these upgrades do is make the diff bigger and add more to read later if someone is investigating a bug.

rmg:
I only had minor style nits/suggestions that can be ignored.. I'll leave the real review to @bajtos.

test/common.js
ritch: They are not all PersistedModels.

rmg: Where they all PersistedModels before?

raymondfeng: Some recent models added to workspace are not PersistedModels.

boot-script/index.js
```diff
@@ -57,7 +57,8 @@ module.exports = yeoman.generators.Base.extend({
   generate: function() {
     var source = this.templatePath(this.type + '.js');
 
-    var targetPath = path.normalize('/server/boot/' + this.name + '.js');
+    // yeoman-generator 0.20.x does'nt like the leading /
```
bajtos: Typo? I believe `doesn't` is the correct spelling.

raymondfeng: Yes, it's typo. I'll fix it.

bajtos:
The code change looks good. I am not familiar with changes introduced in yeoman 0.20, I'll trust you @raymondfeng on this.

Please make sure to manually verify the following scenarios - I remember they had issues which are difficult to discover via unit-tests:
 - `slc loopback` - check that `npm install` is run
 - `slc loopback:model`, enter some properties - check that the property sub-generator was correctly invoked
 - `slc loopback:example` - check that all models, properties, relations, etc. are scaffolded

bajtos:
@raymondfeng also please address the minor comments made by me and @rmg above. Other than that, the patch LGTM.

lib/actions.js
bajtos: +1 fore reducing noise in the commit history

package.json
bajtos: +1 for reducing noise in the commit history

raymondfeng:
verified yo loopback, yo loopback:model and yo loopback:example.

PR #120, opened 2015-11-10
Make sure relation name different from property name

### LINKED ISSUE #118: Do not allow relation name to be the same as property name
The issue https://github.com/strongloop/loopback-datasource-juggler/issues/735 noted that although the generator allows it, a relation name should not be the same as the property name.

The generator should provide an appropriate error message and disallow the choice.

jannyHou:
Still work in progress.
Through this pull request I am trying to fix issue 118 (`https://github.com/strongloop/generator-loopback/issues/118`)
Mainly I added a validate function inside prompts to check the name conflict, all these changes happened in /relation/index.js

Fix #118

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -60,6 +60,26 @@ exports.validateName = function (name) {
 };
 
 /**
+ * Check if relation has a different name from properties
+ * @param {Object} modelDefinition The model which has the relation
+ * @param {String} name The user input
+ * @param {Function(String|Boolean)} Callback to receive the check result.
+ */
+exports.checkRelationName = function (modelDefinition, name, callback) {
+  modelDefinition.properties(function(err, list) {
+    if (err) callback(err);
+    var conflict = list.some(function(property) {
+      return property.name === name;
+    });
+    if (conflict) {
+      callback('Same property name already exists: ' + name);
+    } else {
+      callback(true);
+    }
+  });
+};
+
+/**
  * Get the root command name based on the env var
  * @returns {string}
  */
--- a/relation/index.js
+++ b/relation/index.js
@@ -9,6 +9,7 @@ var ModelRelation = workspace.models.ModelRelation;
 var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 var validateName = helpers.validateName;
+var checkRelationName = helpers.checkRelationName;
 
 module.exports = yeoman.generators.Base.extend({
 
@@ -65,6 +66,7 @@ module.exports = yeoman.generators.Base.extend({
   },
 
   askForParameters: function() {
+    var modelDef = this.modelDefinition;
     var done = this.async();
 
     var modelChoices = this.modelNames.concat({
@@ -108,7 +110,11 @@ module.exports = yeoman.generators.Base.extend({
           }
           return m;
         },
-        validate: validateName
+        validate: function(value) {
+          var isValid = validateName(value);
+          if (isValid !== true) return isValid;
+          return checkRelationName(modelDef, value, this.async());
+        }
       },
       {
         name: 'foreignKey',
--- a/test/helpers.test.js
+++ b/test/helpers.test.js
@@ -3,6 +3,7 @@
 var helpers = require('../lib/helpers');
 var validateAppName = helpers.validateAppName;
 var validateName = helpers.validateName;
+var checkRelationName = helpers.checkRelationName;
 require('chai').should();
 var expect = require('chai').expect;
 
@@ -57,6 +58,21 @@ describe('helpers', function() {
       testValidationRejectsValue(validateName, 'm.prop');
     });
   });
+
+  // test checkRelationName()
+  describe('checkRelationName()', function() {
+    var sampleModelDefinition = new ModelDefinition([
+      {name: 'name', id: 1},
+      {name: 'city', id: 2}
+    ]);
+    it('should accept names with no conflict', function() {
+      testRelationAcceptsValue(sampleModelDefinition, 'myrelation');
+    });
+
+    it('should report errors for duplicate name', function() {
+      testRelationRejectsValue(sampleModelDefinition, 'city');
+    });
+  });
 });
 
 function testValidationAcceptsValue(validationFn, value) {
@@ -65,4 +81,23 @@ function testValidationAcceptsValue(validationFn, value) {
 
 function testValidationRejectsValue(validationFn, value) {
   expect(validationFn(value), value).to.be.a('string');
+}
+
+function testRelationAcceptsValue(modelDefinition, value) {
+  checkRelationName(modelDefinition, value, function(result) {
+    expect(result, value).to.equal(true);
+  });
+}
+
+function testRelationRejectsValue(modelDefinition, value) {
+  checkRelationName(modelDefinition, value, function(result) {
+    expect(result, value).to.be.a('string');
+  });
+}
+
+function ModelDefinition(propertyList) {
+  this.propertyList = propertyList;
+  this.properties = function(callback) {
+    callback(null, this.propertyList);
+  };
 }


relation/index.js
```diff
@@ -108,7 +111,39 @@ module.exports = yeoman.generators.Base.extend({
...
           return m;
         },
-        validate: validateName
+        validate: function(value) {
+
+            var done = this.async();
+
+            //validfn returns false if relationship 
+            //has the same name with properties
+            var validfn = function(value, cb) {
```
bajtos: Please refactor this anonymous function and extract a named method defined outside of `askForParameters`.

relation/index.js
```diff
@@ -108,7 +111,39 @@ module.exports = yeoman.generators.Base.extend({
           }
           return m;
         },
-        validate: validateName
+        validate: function(value) {
+
+            var done = this.async();
+
+            //validfn returns false if relationship
```
bajtos: Nitpick: we add a space after `//`:

```js
// validfn ...
```

relation/index.js
```diff
@@ -108,7 +111,39 @@ module.exports = yeoman.generators.Base.extend({
...
+            //validfn returns false if relationship 
+            //has the same name with properties
+            var validfn = function(value, cb) {
+
+              //set flag isValid
+              var isValid = true;
+           
+              modelDef.properties(function(err, list) {
+                if(err) return done(err);
+                else {
```
bajtos: Style: use curly braces consistently. Since you use them in `else` block, it's better to use them in `if` block too. Plus add a space after `if` keyword.

```js
if (err) {
  return done(err);
} else {
  // ...
}
```

relation/index.js
```diff
@@ -108,7 +111,39 @@ module.exports = yeoman.generators.Base.extend({
...
+                  //check each property name in the list
+                  async.each(list, function(property, callback) {
+                    if(property.name === value) isValid = false;
+                    callback();
+                  }, function(err) {
+                    if(err) return done(err);
+                    cb(null, isValid);
+                  });
```
bajtos: No need to use async when the check is sync. You can simplify this code as follows:

```js
var conflict = list.some(function(property) { return property.name === value; });
cb(null, !conflict);
```

relation/index.js
```diff
@@ -108,7 +111,39 @@ module.exports = yeoman.generators.Base.extend({
...
+                }
+              });
+            };
+            
+            //return the result
+            validfn(value, function(err, result){
+              if(err) return done(err);
+              else done(result && validateName(value));
```
bajtos: IMO it's better to call `validateName` first. The operation is faster than checking naming clashes with properties, therefore the user will receive an error sooner. It will also simplify your code, since you can pass `done` directly as `validFn` callback.

```js
validFn(value, done);
```

BTW shouldn't the last line L143 in your code be `done(null, result && validateName(value))`?

relation/index.js
bajtos: Well, this may be not necessary after you apply refactorings that I suggested below.

bajtos:
@jannyHou thank you for the pull request, you are headed in the right direction! I left you few comments how to improve the code. Please add also some unit-tests to verify the implementation and prevent regressions in the future.

jannyHou:
@bajtos thanks for your detailed comments! I refactored most of my code in the way you suggested, it's organized more concisely now. 
And I tried `done(null, result && validateName(value))`, however that doesn't work, I think it's because done() only takes the return value as its parameter, like the doc shows: https://github.com/sboudrias/Inquirer.js#question
Now I am working on the unit test using chai. Actually windows has some problem running the test in repo, but it doesn't conflict with my test in a very serious way, and next I will also look into why the test error happens.
Regards.

relation/index.js
```diff
@@ -10,6 +10,7 @@ var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 var validateName = helpers.validateName;
 
+
```
bajtos: Extra empty line, please remove.

relation/index.js
```diff
@@ -108,7 +110,25 @@ module.exports = yeoman.generators.Base.extend({
...
+        validate: function(value) {
+
+          var done = this.async();
+
+          if (validateName(value) !== true) {
+            done(validateName(value));
+          } 
+
+          modelDef.properties(function(err, list) {
+            if (err) return done(err);
```
bajtos: Since the `done` callback expects a single argument with a boolean value, I think we should handle the error different way here. Something like

```js
if (err) {
  console.log('Cannot validate relation name against existing model properties:', err.message);
  // Let the user add this relation even if we cannot confirm there is no naming clash
  return done(true);
}
```

relation/index.js
```diff
@@ -108,7 +110,25 @@ module.exports = yeoman.generators.Base.extend({
           }
           return m;
         },
-        validate: validateName
+        validate: function(value) {
+
+          var done = this.async();
+
+          if (validateName(value) !== true) {
+            done(validateName(value));
```
bajtos: Don't call `validateName` twice - either save the result to a variable, or call `done(false)` here on L118.

bajtos: Ooops, I did not realize `done` expects an error string. In that case you should definitely save the result of `validateName` to a variable.

relation/index.js
```diff
@@ -108,7 +110,25 @@ module.exports = yeoman.generators.Base.extend({
...
+          } 
+
+          modelDef.properties(function(err, list) {
+            if (err) return done(err);
+            var conflict = list.some(function(property) { 
+              return property.name === value; 
+            });
+            if (conflict) {
+              done('Same property name already exists: ' + value);
```
bajtos: Missing `return` statement. At the moment, `done` will be called twice.

relation/index.js
```diff
@@ -108,7 +110,25 @@ module.exports = yeoman.generators.Base.extend({
           }
           return m;
         },
-        validate: validateName
+        validate: function(value) {
+
```
bajtos: Extra empty line, please remove.

bajtos:
The code looks much better now! 

> Now I am working on the unit test using chai. Actually windows has some problem running the test in repo, but it doesn't conflict with my test in a very serious way, and next I will also look into why the test error happens.

Sounds good. Feel free to ping me on SlackChat if you can't figure this out in a reasonable time on your own.

jannyHou:
Thanks @bajtos , I modified the code with return and removed empty lines. 
Well I didn't  change `if (err) return done(err);`, because I think `done` takes either error or false and `console.log` might cause format problem on screen, we can discuss it tomorrow:)
I also uploaded my test case, I added a test case 'check naming conflict for relation', and I explained my problem in the comments of that snippet.

bajtos:
> Well I didn't change if (err) return done(err);, because I think done takes either error or false and console.log might cause format problem on screen, we can discuss it tomorrow:)

I checked the current [source code](https://github.com/SBoudrias/Inquirer.js/blob/6d77ea68c352a5a0948c1a95d405d30c4d406b41/lib/prompts/base.js#L91-L114), looks like `err` is treated as a string and printed via `toString`. Which may produce a message that is not very helpful to the user. It will be best to try this in practice to learn how the system really behaves, instead of guessing like we do now. Let's discuss this later today.

> I also uploaded my test case, I added a test case 'check naming conflict for relation', and I explained my problem in the comments of that snippet.

Hmm, let me check the code.

test/relations.test.js
```diff
@@ -31,6 +31,60 @@ describe('loopback:relation generator', function() {
...
+        type: 'boolean',
+        required: true
+      });
+    });
+
+    // name a relationship as 'year', 
+    // should get error here but the validate function doesn't work
+    // even I name it as 'year@@@', it passes the test
+    // validateName doesn'n work either
+    // but they all work fine when I run slc loopback:relation in cmd
```
bajtos: Apparently the mock prompt does not run validations - see [source code](https://github.com/yeoman/generator/blob/a7d8dd0fcaed7a12e440579ecafaf1287d326f1f/lib/test/adapter.js#L12-L41).

I am proposing to extract the validate method into helpers (e.g. `helpers.validateRelationName`) and test it in isolation, like we test `validateName` - see [test/helpers.js](https://github.com/strongloop/generator-loopback/blob/a884c0b27314e3b2602fff1c3b6f75b602e0b57a/test/helpers.test.js).

jannyHou:
Hi @bajtos, these changes are taken after our chat on last Friday, I mainly made two modifications:
1. Extracted the validation function into `lib/helpers.js`, named `checkRelationName`.
2. Created test cases in `helpers.test.js`, mock ModelDefinition in the test file.
Thanks!

lib/helpers.js
```diff
@@ -60,8 +60,29 @@ exports.validateName = function (name) {
 };
 
 /**
+ * Check if relation has a different name from properties
+ * @param {Object} modelDefinition The model which has the relation
+ * @param {String} name The user input
+ * @returns {String|Boolean} The check result
```
bajtos: Since this function is async, you should describe the callback args, not the return value.

```js
/**
 * @param {Function(String|Boolean)} Callback to receive the check result.
 */
```

lib/helpers.js
```diff
@@ -60,8 +60,29 @@ exports.validateName = function (name) {
...
+exports.checkRelationName = function (modelDefinition, name, callback) {
+  modelDefinition.properties(function(err, list) {
+    if (err) callback(err);
+    var conflict = list.some(function(property) { 
+      return property.name === name; 
+    });
+    if (conflict) {
+      callback('Same property name already exists: ' + name);
+    }
+    else {
```
bajtos: Coding style: merge L76 and L77.

```js
} else {
```

lib/helpers.js
```diff
@@ -60,8 +60,29 @@ exports.validateName = function (name) {
...
+      callback(true);
+    }
+  });
+};
+
+/**
  * Get the root command name based on the env var
- * @returns {string}
+ * @returns {string}
```
bajtos: Trailing space, please remove.

Most text editors can be configured to automatically remove trailing spaces, I find it personally best to turn on this option and don't have to worry about trailing spaces at all.

test/helpers.test.js
```diff
@@ -3,10 +3,14 @@
...
 var validateAppName = helpers.validateAppName;
 var validateName = helpers.validateName;
+var checkRelationName = helpers.checkRelationName;
 require('chai').should();
 var expect = require('chai').expect;
 
 describe('helpers', function() {
+  describe('checkRelationName()', function() {
+
+  });
```
bajtos: Looks like a forgotten code, please remove.

test/helpers.test.js
```diff
@@ -57,6 +61,23 @@ describe('helpers', function() {
...
     });
   });
+
+  // test checkRelationName()
+  describe('checkRelationName()', function() {
+    var firstModelDefinition = new ModelDefinition([
+      {name: 'name', id: 1},
+      {name: 'city', id: 2}
+    ]);
+    var secondModelDefinition = new ModelDefinition([]);
```
bajtos: This second model definition looks redundant to me. The algorithm does not depend on the number of model properties, so why is a model with no properties so special that it deserves an explicit test?

test/helpers.test.js
```diff
@@ -66,3 +87,22 @@ function testValidationAcceptsValue(validationFn, value) {
...
+  checkRelationName(modelDefinition, value, function(result) {
+    expect(result, value).to.be.a('string');
+  });
+}
+
+function ModelDefinition(propertyList) {
+  this.propertyList = propertyList;
+  this.properties = function(callback) {
+    callback(null, this.propertyList);
+  };
```
bajtos: We usually don't mock LoopBack models. However, since setting up temporary workspace models is non-trivial, I think it's ok to make an exception here.

jannyHou:
Thanks! em... always check `null` and `empty` is a habit I gained from leetcode, but you are right, the function has only one behaviour no matter haw many properties exist. I will enhance my sublime later!

jannyHou:
Thanks @bajtos, I changed the code according to your suggestion. Can you review it when you have time?

bajtos:
Cool, the patch LGTM now.

Could you please clean up the commit history and squash all commits into a single one with a descriptive commit message following the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)?

If you are not familiar with git history rewriting, then here are the commands you need: 1) `git rebase -i master` 2) `git push -f`.

jannyHou:
@bajtos No problem, I will convert commits to only one and....wait until we are approved :) Thanks a lot!

bajtos:
> wait until we are approved :)

Cool. Please ping me once that happens so that we can get this landed.

bajtos:
Landed, thank you for the contribution :)

PR #121, opened 2015-11-23
Add command line printing swagger definition to loopback

0candy:
Second part of the change for command line printing for swagger in loopback.

@bajtos Please review.

Connect to strongloop/loopback#1783

### DIFF
--- a/export-api-def/USAGE
+++ b/export-api-def/USAGE
@@ -1,0 +1,14 @@
+Description:
+  Generate Swagger API definitions for LoopBack applications.
+
+Example:
+
+  slc loopback:export-api-def [--json]
+
+  This will output the api definitions to stdout.
+  Use option --json to display in JSON format, otherwise it defaults to YAML.
+
+  slc loopback:export-api-def [--o outputFile]
+
+  This will output the api definitions to the specified location. Based on the
+  file extension, it will print to yaml or json format.
--- a/export-api-def/index.js
+++ b/export-api-def/index.js
@@ -1,0 +1,41 @@
+'use strict';
+var fs = require('fs');
+var path = require('path');
+var yeoman = require('yeoman-generator');
+var helpers = require('../lib/helpers');
+var apiGenerator = require('loopback-api-definition');
+var mkdirp = require('mkdirp');
+
+module.exports = yeoman.generators.Base.extend({
+
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+    var format = this.options.json;
+    this.format = format ? 'json' : 'yaml';
+    this.option('output', {
+      type: String,
+      alias: 'o',
+      desc: 'Name/Full path to the output file.',
+      defaults: ''
+    });
+  },
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  generate: function() {
+    var filePath = this.options.output;
+    var options = { format: this.format, output: filePath };
+    var app = require(this.destinationRoot());
+    var apiDef = apiGenerator.getApiDef(app, options);
+
+    // Print to console if no output file specified.
+    if (filePath) {
+      mkdirp.sync(path.dirname(filePath));
+      fs.writeFileSync(filePath, apiDef);
+    } else {
+      process.stdout.write(apiDef);
+    }
+  }
+});
--- a/package.json
+++ b/package.json
@@ -28,8 +28,10 @@
     "debug": "^2.2.0",
     "inflection": "^1.7.1",
     "js-yaml": "^3.4.2",
+    "loopback-api-definition": "^1.0.0",
     "loopback-swagger": "^2.0.0",
     "loopback-workspace": "^3.18.0",
+    "mkdirp": "^0.5.1",
     "request": "^2.60.0",
     "yeoman-generator": "^0.21.1",
     "yosay": "^1.0.5"
--- a/test/export-api-def.test.js
+++ b/test/export-api-def.test.js
@@ -1,0 +1,119 @@
+'use strict';
+var path = require('path');
+var helpers = require('yeoman-generator').test;
+var fs = require('fs');
+var expect = require('chai').expect;
+var wsModels = require('loopback-workspace').models;
+var common = require('./common');
+var yaml = require('js-yaml');
+var install = require('strong-cached-install');
+var SANDBOX =  path.resolve(__dirname, 'sandbox');
+var PKG_CACHE = path.resolve(__dirname, '..', '.pkgcache');
+
+describe('loopback:export-api-def generator', function() {
+  this.timeout(30 * 60 * 1000); // 30 minutes
+  before(common.resetWorkspace);
+
+  before(function createSandbox(done) {
+    helpers.testDirectory(SANDBOX, done);
+  });
+
+  before(function createProject(done) {
+    common.createDummyProject(SANDBOX, 'test-app', done);
+  });
+
+  before(function createProductModel(done) {
+    var test = this;
+    wsModels.ModelDefinition.create(
+      {
+        name: 'Product',
+        facetName: 'common'
+      },
+      function(err, model) {
+        test.Model = model;
+        done(err);
+      });
+  });
+
+  before(function installDependencies(done) {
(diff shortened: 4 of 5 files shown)


export-api-def/index.js
```diff
@@ -0,0 +1,21 @@
...
+    var format = this.options.json;
+    this.format = format ? 'json' : 'yaml';
+  },
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  generate: function() {
+    process.stdout.write(apigenerator.getApiDef(require(process.cwd()), {format:this.format}));
```
bajtos: There is so much being done on this single line. Would you mind splitting it into several lines, one step per line?

export-api-def/USAGE
```diff
@@ -0,0 +1,10 @@
+Description:
+  Generate Swagger API definitions for LoopBack application.
+
+Example:
+
+  slc loopback:export-api-def [--json]
+
+  This will output the api definitions to stdout.
+  Use option --json to display in JSON format, otherwise it defaults to YAML.
+  Output can be redirected to a file by adding > output.yaml
```
bajtos: It would be nice to have `-o outfile` so that users don't have to rely on output redirection. It will also allow us to trigger the file-based format detection which is already implemented by loopback-api-definition.

bajtos:
@0candy thank you for the update. I am missing unit-tests for this new generator, is there any good reason against writing them?

0candy:
DO NOT MERGE this change until https://github.com/strongloop/loopback-api-definition modules gets approved and published.

export-api-def/USAGE
```diff
@@ -0,0 +1,14 @@
+Description:
+  Generate Swagger API definitions for LoopBack application.
```
superkhau: applications

export-api-def/index.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = yeoman.generators.Base.extend({
+
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+    var format = this.options.json;
+    this.format = format ? 'json' : 'yaml';
+    this.option('output', {
+      type: String,
+      alias: 'o',
+      desc: 'Name/location of output file.',
```
superkhau: is the "Full path to the output file"?

export-api-def/index.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+var fs = require('fs');
+var path = require('path');
+var yeoman = require('yeoman-generator');
+var helpers = require('../lib/helpers');
+var apigenerator = require('loopback-api-definition');
```
superkhau: apiGenerator?

export-api-def/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  generate: function() {
+    var filePath = this.options.output;
+    var options = { format: this.format, output: filePath };
+    var app = require(this.destinationRoot());
+    var apiDef = apigenerator.getApiDef(app, options);
```
superkhau: apiGenerator

test/exportapi.test.js
```diff
@@ -0,0 +1,181 @@
...
+    var content = fs.readFileSync(jsonFile, 'utf-8');
+    return JSON.parse(content);
+  }
+
+  function readYamlFile() {
+    var yamlFile = path.resolve(SANDBOX, 'swagger.yaml');
+    var content = fs.readFileSync(yamlFile, 'utf-8');
+    return yaml.load(content);
+  }
+
```
superkhau: remove space

superkhau:
@0candy Minor nitpicks, otherwise LGTM. @bajtos for final review and OK to merge.

package.json
```diff
@@ -36,6 +37,7 @@
   },
   "devDependencies": {
     "chai": "^3.2.0",
+    "compression": "^1.6.0",
```
bajtos: @0candy why is `compression` needed in dev dependencies?

test/exportapi.test.js
```diff
@@ -0,0 +1,180 @@
...
+var path = require('path');
+var helpers = require('yeoman-generator').test;
+var SANDBOX =  path.resolve(__dirname, 'sandbox');
+var fs = require('fs');
+var expect = require('chai').expect;
+var wsModels = require('loopback-workspace').models;
+var common = require('./common');
+var yaml = require('js-yaml');
+
+describe('loopback:model generator', function() {
```
bajtos: I think this should be called `loopback:export-api-def` instead.

Also please rename the test file to `export-api-def.test.js` so that it matches the generator name.

test/exportapi.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+  beforeEach(function addRestDataSource(done) {
+    wsModels.DataSourceDefinition.create({
+      name: 'rest',
+      connector: 'rest',
+      facetName: 'server'
+    }, done);
+  });
+
+  it('creates common/models/{name}.json', function(done) {
```
bajtos: `loopback:export-api-def creates common/models/{name}.json` - that does not sound right. If the tests for `export-api-def` need more models to be created, then create then in `beforeEach`  hook(s), not in a test case.

```js
beforeEach(function createProductModel(done) {
  var modelGen = givenModelGenerator();
  // etc.
});
```

Since we are not testing the model generator, it's better to use `wsModels.ModelDefinition` API directly.

package.json
bajtos: I think you need to modify your test to install dependencies in the scaffolded application.

```js
var install = require('strong-cached-install');
var PKG_CACHE = path.resolve(__dirname, '..', '.pkgcache');
// ...

beforeEach(function installDependencies(done) {
  install(SANDBOX, PKG_CACHE, ['dependencies'], done);
});
```

bajtos: see https://github.com/strongloop/generator-loopback/blob/17e01ccfcf1abd8f37361ee985412ce05e5475c8/test/end-to-end/example.e2e.js

0candy:
@bajtos Please review

package.json
```diff
@@ -28,6 +28,7 @@
     "debug": "^2.2.0",
     "inflection": "^1.7.1",
     "js-yaml": "^3.4.2",
+    "loopback-api-definition": "1.0.0",
```
bajtos: Please use `^1.0.0`, that way we don't have to change this value every time we publish a new backwards-compatible version of loopback-api-definition.

test/export-api-def.test.js
```diff
@@ -0,0 +1,116 @@
...
+        test.Model = model;
+        done(err);
+      });
+  });
+
+  before(function installDependencies(done) {
+    install(SANDBOX, PKG_CACHE, ['dependencies'], done);
+  });
+
+  it('JSON format', function(done) {
```
bajtos: `it('produces JSON format')` would be a better test name. Same for YAML test below.

test/export-api-def.test.js
```diff
@@ -0,0 +1,116 @@
...
+      done();
+    });
+  });
+
+  it('YAML format', function(done) {
+    var gen = givenGenerator();
+    gen.options.output = 'swagger.yaml';
+
+    gen.run(function() {
+      var content = readYamlFile();
```
bajtos: Nitpick: since YAML is a superset of JSON, this would pass even if we were producing only JSON files. Makes me wonder if there is any reasonably easy way how to check that the file actually uses YAML features that are not in JSON?

bajtos:
@0candy left few minor comments, reassigning the PR to you. Please reassign it back to me when it's ready for another round of review.

0candy:
@bajtos  Made the changes please review.

bajtos:
@slnode test please

bajtos:
@0candy the code looks good now, but the tests are failing:

```
1) loopback:export-api-def generator produces JSON format:
     Error: Cannot find module 'mkdirp'
```

Could you please check? I assume it's just a missing (dev) dependency. If that's the case, then you can land this patch after you fix the build, no need for further review/approval.

0candy:
@slnode test please

Issue #94, opened 2015-03-05
Add remote method generator


superkhau:
We should add the `slc loopback:remote-method` command to scaffold custom remote methods.

See #90 

**Subtasks**
- [x] First cut implementation
- [x] Http mapping of arguments (see https://docs.strongloop.com/display/public/LB/Remote+methods?src=search&src=search#Remotemethods-HTTPmappingofinputarguments)
- [x] Add `isStatic`
- [x] Check input valid
- [x] Add unit tests
- [x] Either add a dummy method implementation to model js file or alternatively print instructions including dummy method code to the output after the JSON metadata was updated.
- [x] Add http verb(major and custom)
- [x] Make modelName an argument

bajtos:
It would be good to land https://github.com/strongloop/loopback/pull/1163 first.

jannyHou:
A record for subtasks:
- [x] first cut implementation
- [ ] http mapping of arguments
- [ ] add `isStatic`
- [ ] check input valid

bajtos:
@jannyHou I moved the list of subtasks to issue description to keep them at the top.

bajtos:
@jannyHou Nice start in #124! Please continue with adding tests for the new generator as it is implemented now before working on additional features.

jannyHou:
@bajtos yeah good to have them on top, thanks a lot! I would add tests first, then move to features.

bajtos:
@jannyHou can we close this issue now, since #124 has been landed?

jannyHou:
@bajtos sure, thank your for reviewing the pr and guiding me to improve the implementation! since it's landed I will close the issue here.

PR #124, opened 2015-11-26
Remote method generator

jannyHou:
hey @bajtos, this is the first implementation of `remote-method generator` 
Further features are recorded in this issue: https://github.com/strongloop/generator-loopback/issues/94
Please review it when you have time. And I will add new fields and features in the near future.
Thanks a lot.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -3,6 +3,7 @@ var assert = require('assert');
 var chalk = require('chalk');
 var yeoman = require('yeoman-generator');
 var Workspace = require('loopback-workspace').models.Workspace;
+var ModelProperty = require('loopback-workspace').models.ModelProperty;
 
 exports.reportValidationError = function(err, logFn) {
   if (!err || err.name !== 'ValidationError') {
@@ -131,3 +132,14 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
   return connectorMeta && connectorMeta.baseModel;
 };
 
+/**
+ * Get the available type choices that should be used for arguments
+ * from workspace.
+ */
+exports.getTypeChoices = function() {
+  var typeChoices = ModelProperty.availableTypes.concat({
+    name: '(other)',
+    value: null
+  });
+  return typeChoices;
+};
--- a/remote-method/USAGE
+++ b/remote-method/USAGE
@@ -1,0 +1,6 @@
+Description:
+  Creates a new Remote Method in the LoopBack application.
+
+Example:
+
+  slc loopback:remote-method
--- a/remote-method/index.js
+++ b/remote-method/index.js
@@ -1,0 +1,460 @@
+'use strict';
+var yeoman = require('yeoman-generator');
+var chalk = require('chalk');
+var util = require('util');
+var path = require('path');
+
+var actions = require('../lib/actions');
+var helpers = require('../lib/helpers');
+var validateName = helpers.validateName;
+var typeChoices = helpers.getTypeChoices;
+var ModelDefinition = require('loopback-workspace').models.ModelDefinition;
+
+module.exports = yeoman.generators.Base.extend({
+  // This generator does not track file changes via yeoman,
+  // as loopback-workspace is editing (modifying) files when
+  // saving project changes.
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+
+    this.argument('modelName', {
+      desc: 'Name of the model',
+      required: false,
+      type: String
+    });
+
+    this.argument('methodName', {
+      desc: 'Name of the remote method',
+      required: false,
+      type: String
+    });
+  },
+
+  help: function() {
+    return helpers.customHelp(this);
+  },
+
+  loadProject: actions.loadProject,
+
+  loadModels: actions.loadModels,
+
+  askForModel: function() {
+    if (this.modelName) {
+      this.modelDefinition = this.projectModels.filter(function(m) {
+        return m.name === this.modelName;
+      }.bind(this))[0];
+
+      if (!this.modelDefinition) {
+        var msg = 'Model not found: ' + this.modelName +
+                  '. Please choose from Model List:';
+        this.log(chalk.red(msg));
+      }
+    }
+
+    if (!this.modelDefinition) {
+      var done = this.async();
+      var prompts = [
+        {
+          name: 'model',
+          message: 'Select the model:',
+          type: 'list',
+          choices: this.modelNames
+        }
+      ];
+      this.prompt(prompts, function(answers) {
+        this.modelName = answers.model;
+        done();
+      }.bind(this));
+    }
+  },
+
+  findModelDefinition: function() {
+    this.modelDefinition = this.projectModels.filter(function(m) {
+      return m.name === this.modelName;
+    }.bind(this))[0];
+
+    if (!this.modelDefinition) {
+      var msg = 'Model not found: ' + this.modelName;
+      this.log(chalk.red(msg));
+      this.async()(new Error(msg));
+    }
+  },
+
+  askForParameters: function() {
+    var done = this.async();
+    var name = this.methodName;
+    var prompts = [
+      {
+        name: 'methodName',
+        message: 'Enter the remote method name:',
+        required: true,
+        default: name,
+        validate: validateName
+      },
+      {
+        name: 'isStatic',
+        message: 'Is Static?',
+        required:true,
+        type: 'confirm',
+        default: true
+      },
+      {
+        name: 'description',
+        message: 'Description for method:'
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      this.methodName = answers.methodName;
+      this.isStatic = answers.isStatic;
+      this.description = answers.description;
+      done();
+    }.bind(this));
+  },
+
+  delimEndpoints: function() {
+    this.log('Let\'s add some HTTP Endpoints now.\n');
+    this.http = [];
+  },
+
+  askForEndpoints: function() {
+    var done = this.async();
+    this.log('Enter an empty path when done.');
+
+    var prompts = [
+      {
+        name: 'httpPath',
+        message: 'Enter the path of this endpoint:',
+        required: true
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      if (answers.httpPath == null || answers.httpPath === '') {
+        return done();
+      }
+      var subprompts = [
+        {
+          name: 'httpVerb',
+          message: 'HTTP verb:',
+          type: 'list',
+          choices: ['post', 'put', 'delete', 'get', {
+            name: '(other)',
+            value: null
+          }]
+        },
+        {
+          name: 'customHttpVerb',
+          message:
+          'Enter the custom http verb',
+          validate: validateName,
+          when: function(answers) {
+            return answers.httpVerb === null;
+          }
+        }
+      ];
+
+      var httpPath = answers.httpPath;
+
+      this.prompt(subprompts, function(answers) {
+        this.http.push({
+          httpPath: httpPath,
+          httpVerb: answers.httpVerb || answers.customHttpVerb
+        });
+        this.log('\nLet\'s add another endpoint.');
+        this.askForEndpoints();
+      }.bind(this));
+    }.bind(this));
+  },
+
+  delimAccepts: function() {
+    this.log('Let\'s add some Accept Arguments now.\n');
+    this.accepts = [];
+  },
+
+  askForAccepts: function() {
+    var done = this.async();
+    this.log('Enter an empty name when done.');
+
+    var prompts = [
+      {
+        name: 'acceptsArg',
+        message: 'Enter name for accept argument:',
+        required: true,
+        validate: validateName
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      if (answers.acceptsArg == null || answers.acceptsArg === '') {
+        return done();
+      }
+      var subprompts = [
+        {
+          name: 'acceptsType',
+          message: 'The type of accept argument:',
+          type: 'list',
+          choices: typeChoices
+        },
+        {
+          name: 'acceptsRequired',
+          message: 'Required?',
+          type: 'confirm',
+          default: false
+        },
+        {
+          name: 'acceptsDes',
+          message: 'Description of accept argument:',
+          required: true
+        },
+        {
+          name: 'httpSource',
+          message: 'Mapping to http source:',
+          type: 'list',
+          choices: [
+            {
+              name: '(auto)',
+              value: null
+            },
+            'form',
+            'path',
+            'query',
+            {
+              name: 'store the whole request body',
+              value: 'body'
+            },
+            {
+              name: 'store the full request object',
+              value: 'req'
+            },
+            {
+              name: 'store the full response object',
+              value: 'res'
+            },
+            {
+              name: 'store the whole context object',
+              value: 'context'
+            }
+          ]
+        }
+      ];
+
+      var argName = answers.acceptsArg;
+
+      this.prompt(subprompts, function(answers) {
+        var entry = {
+          arg: argName,
+          type: answers.acceptsType,
+          required: answers.acceptsRequired,
+          description: answers.acceptsDes,
+        };
+        if (answers.httpSource) {
+          entry.http = { source: answers.httpSource };
+        }
+        this.accepts.push(entry);
+        this.log('\nLet\'s add another accept argument.');
+        this.askForAccepts();
+      }.bind(this));
+    }.bind(this));
+  },
+
+  delimReturns: function() {
+    this.log('Let\'s add some Return Arguments now.\n');
+    this.returns = [];
+  },
+
+  askForReturns: function() {
+    var done = this.async();
+    this.log('Enter an empty name when done.');
+
+    var prompts = [
+      {
+        name: 'returnsArg',
+        message: 'Enter name for return argument:',
+        required: true,
+        validate: validateName
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      if (answers.returnsArg == null || answers.returnsArg === '') {
+        return done();
+      }
+      var subprompts = [
+        {
+          name: 'returnsType',
+          message: 'The type of return argument:',
+          type: 'list',
+          choices: typeChoices
+        },
+        {
+          name: 'returnsRoot',
+          message: 'Is it root',
+          type: 'confirm',
+          default: false
+        },
+        {
+          name: 'returnsDes',
+          message: 'Description of return argument:',
+          required: true
+        }
+      ];
+
+      var argName = answers.returnsArg;
+
+      this.prompt(subprompts, function(answers) {
+        this.returns.push({
+          arg: argName,
+          type: answers.returnsType,
+          root: answers.returnsRoot,
+          description: answers.returnsDes
+        });
+        this.log('\nLet\'s add another return argument.');
+        this.askForReturns();
+      }.bind(this));
+    }.bind(this));
+  },
+
+  remote: function() {
+    var done = this.async();
+    var def = {
+      name: this.methodName,
+      isStatic: this.isStatic,
+      description: this.description,
+      accepts: this.accepts,
+      returns: this.returns,
+      http: this.http
+    };
+
+    this.modelDefinition.methods.create(def, function(err) {
+      helpers.reportValidationError(err, this.log);
+      return done(err);
+    }.bind(this));
+  },
+
+  printSampleRemoteMethodSource: function() {
+    var text = [
+      buildIntroduction(this),
+      buildJsdoc(this),
+      buildMethodSource(this)
+    ].join('\n\n');
+    this.log(text);
+    // print an empty line as a visual delimiter
+    this.log();
+  },
+
+  saveProject: actions.saveProject
+});
+
+function buildIntroduction(def) {
+  var modelDef = def.modelDefinition;
+  var projectRoot = def.destinationRoot();
+  var jsonFilePath = ModelDefinition.getPath(modelDef.facetName, modelDef);
+  var jsFilePath = modelDef.getScriptPath();
+  jsFilePath = path.relative(projectRoot, jsFilePath);
+  jsFilePath = chalk.yellow(jsFilePath);
+  jsonFilePath = chalk.yellow(jsonFilePath);
+
+  var tip = [
+    util.format(
+      'We have added strong-remoting metadata for your new method to %s',
+      jsonFilePath
+    ),
+    util.format(
+      'Now it\'s up to you to provide implementation in %s',
+      jsFilePath
+    ),
+    'Here is a sample code to get you started:'
+  ].join('\n');
+
+  return tip;
+}
+
+function buildJsdoc(def) {
+  var jsdocLines =  [
+    '/**',
+    util.format(' * %s', def.description),
+  ];
+
+  jsdocLines = jsdocLines.concat(buildInputJsdoc(def.accepts));
+  jsdocLines = jsdocLines.concat(buildReturnsJsdoc(def.returns));
+
+  jsdocLines.push(' */');
+  return jsdocLines.join('\n');
+}
+
+function buildMethodSource(def) {
+  var ref = util.format(
+    '%s%s%s',
+    def.modelName,
+    (def.isStatic ? '.' : '.prototype.'),
+    def.methodName
+  );
+  var functionDef = [util.format(
+    ref + ' = funtion(%s) {',
+    chalk.green(buildInputArgs(def.accepts).join(', '))
+    )
+  ];
+
+  if (def.returns.length > 0)
+    functionDef.push(buildReturnDeclare(def.returns));
+
+  functionDef.push('  // TODO');
+  functionDef.push('  ' + buildCallback(def.returns));
+  functionDef.push('};');
+
+  return functionDef.join('\n');
+}
+
+function buildInputJsdoc(accepts) {
+  var acceptJsdoc = accepts.map(function(accept) {
+    return util.format(
+      ' * \@param {%s} %s %s',
+      accept.type,
+      chalk.green(accept.arg),
+      accept.description
+    );
+  });
+
+  return acceptJsdoc;
+}
+
+function buildReturnsJsdoc(returns) {
+  var returnTypes = ['Error'].concat(returns.map(function(_return) {
+    return _return.type;
+  })).join(', ');
+  var returnDef = util.format(
+    ' * \@param {Function(%s)} callback',
+    chalk.green(returnTypes)
+  );
+  return returnDef;
+}
+
+function buildInputArgs(accepts) {
+  var acceptArg = accepts.map(function(accept) {
+    return accept.arg;
+  });
+  acceptArg.push('callback');
+  return acceptArg;
+}
+
+function buildReturnDeclare(returns) {
+  var returnArgs = buildReturnArgs(returns);
+  returnArgs.shift();
+  var returnDef = util.format(
+    '  var %s;',
+    chalk.green(returnArgs.join(', '))
+  );
+  return returnDef;
+}
+
+function buildCallback(returns) {
+  var callbackLine = util.format(
+    'callback(%s);',
+    chalk.green(buildReturnArgs(returns).join(', '))
+  );
+  return callbackLine;
+}
+
+function buildReturnArgs(returns) {
+  var returnArg = ['null'].concat(returns.map(function(_return) {
+    return _return.arg;
+  }));
+  return returnArg;
+}
--- a/test/remote-method.test.js
+++ b/test/remote-method.test.js
@@ -1,0 +1,65 @@
+/*global describe, beforeEach, it */
+'use strict';
+var path = require('path');
+var helpers = require('yeoman-generator').test;
+var wsModels = require('loopback-workspace').models;
+var SANDBOX =  path.resolve(__dirname, 'sandbox');
+var expect = require('chai').expect;
+var common = require('./common');
+
+describe('loopback:remote-method generator', function() {
+  beforeEach(common.resetWorkspace);
+  beforeEach(function createSandbox(done) {
+    helpers.testDirectory(SANDBOX, done);
+  });
+
+  beforeEach(function createProject(done) {
+    common.createDummyProject(SANDBOX, 'test-app', done);
+  });
+
+  beforeEach(function createCarModel(done) {
+    var test = this;
+    wsModels.ModelDefinition.create(
+      {
+        name: 'Car',
+        facetName: 'common'
+      },
+      function(err, model) {
+        test.Model = model;
+        done(err);
+      });
+  });
+
+  it('adds an entry to common/models/{name}.json', function(done) {
+    var methodGenerator = givenMethodGenerator();
+    helpers.mockPrompt(methodGenerator, {
+      model: 'Car',
+      methodName: 'myRemote',
+      isStatic: 'true',
+      desription: 'This is my first remote method',
+      httpPath: '',
+      acceptsArg: '',
+      returnsArg: ''
+    });
+
+    methodGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var methods = definition.methods || {};
+      expect(methods).to.have.property('myRemote');
+      expect(methods.myRemote).to.eql({
+        isStatic: true,
+        accepts: [],
+        returns: [],
+        http: []
+      });
+      done();
+    });
+  });
+
+  function givenMethodGenerator() {
+    var name = 'loopback:remote-method';
+    var path = '../../remote-method';
+    var gen = common.createGenerator(name, path);
+    return gen;
+  }
+});
--- a/test/common.js
+++ b/test/common.js
@@ -2,6 +2,7 @@
 var async = require('async');
 var fs = require('fs');
 var path = require('path');
+var SANDBOX =  path.resolve(__dirname, 'sandbox');
 var generators = require('yeoman-generator');
 var workspace = require('loopback-workspace');
 var Workspace = workspace.models.Workspace;
@@ -61,4 +62,10 @@ exports.createExampleGenerator = function() {
   };
 
   return exports.createGenerator(name, path, deps, args, options);
+};
+
+exports.readJsonSync = function(relativePath) {
+    var filepath = path.resolve(SANDBOX, relativePath);
+    var content = fs.readFileSync(filepath, 'utf-8');
+    return JSON.parse(content);
 };


remote-method/index.js
```diff
@@ -0,0 +1,244 @@
...
+var ModelProperty = workspace.models.ModelProperty;
+
+var actions = require('../lib/actions');
+var helpers = require('../lib/helpers');
+var validateName = helpers.validateName;
+var accepts = [];
+var returns = [];
+
+module.exports = yeoman.generators.Base.extend({
+  // NOTE(bajtos)
```
bajtos: Please remove my name from the `NOTE` since it's not me who added it ;-) It's ok to keep the note itself though.

remote-method/index.js
```diff
@@ -0,0 +1,244 @@
...
+    return helpers.customHelp(this);
+  },
+
+  loadProject: actions.loadProject,
+
+  loadModels: actions.loadModels,
+
+  askForModel: function() {
+    if (this.options.modelName) {
+      this.modelName = this.options.modelName;
```
bajtos: I think you need to declare `options.modelName` in the constructor function to have it described in help. Could you please check?

See https://github.com/strongloop/generator-loopback/blob/99c66a66d4d1e1179749d792ac3c298fd825636c/app/index.js#L20-L29

BTW I think it makes more sense to make the name an argument. Below is the difference between options and arguments as I understand them (please check it yourself, I may be wrong!):

```
// argument
slc loopback:remote-method MyModel
// option
slc loopback:remote-method --modelName MyModel
```

remote-method/USAGE
```diff
@@ -0,0 +1,6 @@
+Description:
+  Creates a new remote method in the LoopBack application.
+
+Example:
+
+  slc loopback:remote-method
```
bajtos: > `slc loopback:remote-method`

I am proposing to shorten this to `slc loopback:method` to make the command shorter to type.

@ritch @superkhau thoughts?

remote-method/index.js
```diff
@@ -0,0 +1,244 @@
...
+      {
+        name: 'httpPath',
+        message: 'HTTP path:',
+        required: true
+      },
+      {
+        name: 'httpVerb',
+        message: 'HTTP verb:',
+        type: 'list',
+        choices: ['post', 'put', 'delete', 'get']
```
bajtos: > `choices: ['post', 'put', 'delete', 'get']`

Please add at least `head` and `patch`.

Here is a full list of methods as reported by `http.METHODS` in node@4:

```
> http.METHODS
[ 'CHECKOUT',
  'CONNECT',
  'COPY',
  'DELETE',
  'GET',
  'HEAD',
  'LOCK',
  'M-SEARCH',
  'MERGE',
  'MKACTIVITY',
  'MKCALENDAR',
  'MKCOL',
  'MOVE',
  'NOTIFY',
  'OPTIONS',
  'PATCH',
  'POST',
  'PROPFIND',
  'PROPPATCH',
  'PURGE',
  'PUT',
  'REPORT',
  'SEARCH',
  'SUBSCRIBE',
  'TRACE',
  'UNLOCK',
  'UNSUBSCRIBE' ]
```

It makes me wonder whether we should allow users to enter a custom method?

@superkhau @ritch @raymondfeng thoughts?

remote-method/index.js
```diff
@@ -0,0 +1,244 @@
...
+          name: 'returnsType',
+          message: 'The type of return argument:',
+          type: 'list',
+          choices: typeChoices
+        },
+        {
+          name: 'returnsRoot',
+          message: 'Is it root',
+          type: 'confirm',
+          default: true
```
bajtos: I think it's better to use `default:false` here.

remote-method/index.js
```diff
@@ -0,0 +1,244 @@
...
+    this.log('Let\'s add some Accept Arguments now.\n');
+  },
+
+  askForAccepts: function() {
+    var done = this.async();
+
+    var typeChoices = ModelProperty.availableTypes.concat({
+      name: '(other)',
+      value: null
+    });
```
bajtos: This block is repeated twice, please extract it to shared constant or shared helper function.

bajtos:
Thank you @jannyHou for the pull request. The implementation looks mostly good! I need you to add tests before this can be landed.

jannyHou:
thanks @bajtos ! I am working on test cases now and would commit them with some trivial improvements together in the near future.

remote-method/USAGE
superkhau: Can we alias the command? ie. `slc loopback:remote-method` calls `slc loopback:method`, best to have both IMO. We can document `slc loopback:remote-method` as the official way and for experts, they can look up the shortcuts.

remote-method/index.js
superkhau: +1 on supporting only the major 6 and leaving the rest to user customization.

remote-method/index.js
superkhau: I think we should come up with a convention for targeting people in comments: ie @bajtos like GH to make it obvious.

remote-method/USAGE
bajtos: @superkhau what are your concerns regarding `slc loopback:method`? What other kinds of methods with a generator do you envision?

I'd rather avoid maintaining aliases.

jannyHou:
hey @bajtos, I added testing and made some changes to code.
1. In the testing, accepts and returns are set to empty, I am still thinking about how to do the loop when mock the generator.
2. I extracted the `typeChoices` to be returned from `helpers.js`
3. I set `methodName` to be the argument and temporarily canceled `option.modelName`, so now when you run
```
slc loopback:remote-method myMethodName
```
the default methodName would be `myMethodName`. If the option is still need to set, I think I d better to confirm something with you tmr:)
4. Two major method types are added, I would work on custom types later.

lib/helpers.js
```diff
@@ -3,6 +3,8 @@ var assert = require('assert');
 var chalk = require('chalk');
 var yeoman = require('yeoman-generator');
 var Workspace = require('loopback-workspace').models.Workspace;
+var workspace = require('loopback-workspace');
```
bajtos: please use the new `workspace` property in L5 above too.

remote-method/USAGE
bajtos: > what are your concerns regarding slc loopback:method? What other kinds of methods with a generator do you envision?
> I'd rather avoid maintaining aliases.

@superkhau ping

bajtos:
> I set `methodName` to be the argument and temporarily canceled `option.modelName`, so now when you run
```
slc loopback:remote-method myMethodName
```
the default methodName would be `myMethodName`. If the option is still need to set, I think I d better to confirm something with you tmr:)

I find that a bit unusual that one can set method name as an argument but cannot set the model name.

How about `slc loopback:remote-method MyModel myMethodName`?

If we want to get fancy, then we could even support dot-based notation, but I think that's probably not worth our effort.

```
slc loopback:remote-method MyModel.myMethodName
slc loopback:remote-method MyModel.prototype.myMethodName
```

test/remote-method.test.js
```diff
@@ -0,0 +1,76 @@
...
+      desription: 'This is my first remote method',
+      httpPath: '/myRemote',
+      httpVerb: 'post',
+      acceptsArg: '',
+      returnsArg: ''
+    });
+
+    methodGenerator.run(function() {
+      var definition = readJsonSync('common/models/car.json');
+      var meths = definition.methods || {};
```
bajtos: I find `meths` rather confusing, please use the full name `methods`.

test/remote-method.test.js
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  function givenMethodGenerator() {
+    var name = 'loopback:remote-method';
+    var path = '../../remote-method';
+    var gen = common.createGenerator(name, path);
+    return gen;
+  }
+
+  function readJsonSync(relativePath) {
```
bajtos: This function is duplicated in several test files. Please extract it to `test/common.js`, e.g. `common.readSandboxJsonSync(relativePath)`.

Ideally, this refactoring should be a standalone commit, the first one if this patch (use `git rebase -i master && git push -f`). Alternatively, you can send this refactoring in a new pull request that we can quickly land so that you can rebase this PR on top of that change. Please let me know what would be the easiest way for your.

remote-method/index.js
```diff
@@ -0,0 +1,249 @@
...
+        }
+      }
+    ];
+    this.prompt(prompts, function(answers) {
+      this.methodName = answers.methodName;
+      this.isStatic = answers.isStatic;
+      this.http = {
+        path: answers.httpPath,
+        verb: answers.httpVerb
+      };
```
bajtos: FWIW, one method may be mapped to multiple HTTP endpoints - see https://github.com/strongloop/loopback-workspace/pull/237 and a real-world example in [loopback/lib/persisted-model.js](https://github.com/strongloop/loopback/blob/bfd5059d119d7a7131a04c9873cdc61f04aa1165/lib/persisted-model.js#L560-L563).

For the first version, I think it's ok to support only one http endpoint in the generator. I.e. you can keep `httpPath` and `httpVerb`, but make sure you create `this.http` as an array here.

bajtos:
> In the testing, accepts and returns are set to empty, I am still thinking about how to do the loop when mock the generator.

I think that would require you to write a custom mock of Yeoman's prompt object. I'm ok to leave this out of this PR (not covered by tests), we don't test looping questions in the model generator either.

>  I extracted the typeChoices to be returned from helpers.js

:+1:

remote-method/index.js
```diff
@@ -0,0 +1,249 @@
...
+      {
+        name: 'httpPath',
+        message: 'HTTP path:',
+        required: true
+      },
+      {
+        name: 'httpVerb',
+        message: 'HTTP verb:',
+        type: 'list',
+        choices: ['post', 'put', 'delete', 'get', 'head', 'patch', 'other']
```
bajtos: We use `null` for custom value in other places (see e.g. `getTypeChoices`), please stay consistent and use the same approach here too.

```js
choices: [
  'post', 'put', 'delete', 'get', 'head', 'patch',
  { name: '(other)', value: null }
 ]
```

test/remote-method.test.js
jannyHou: yes it's wired, I changed it.

test/remote-method.test.js
jannyHou: hmm...actually `typeChoices` also repeats in several generator files, I prefer to apply it and `readJsonSync` together after this pr :)

remote-method/index.js
jannyHou: I think it's not that complex to turn it an array, same as the loop of accepts and returns, I will implement it as an array in next commit.

jannyHou:
Hey @bajtos,  I applied your comments and here is the major changes I made today:
1. Turn http to be an array, create a loop input as `accepts` and `returns`.
2. Make `modelName` as an argument. My expected user cases are: 
(a). When `modelName` is provided in cmd, user doesn't need to type in again, the generator skips `askForModel` and go to next prompt.
(b). When no argument is provided in cmd, generator asks for `modelName`.
(c). When user provides a wrong `modelName` in argument, warns him 'Model not found: ' + this.modelName + '. Please choose from Model List:', and generator asks for `modelName`
3. Change `others` to `{ name: '(other)', value: null }` to keep it consistent.
4. Extract `readJsonSync()` to `test/common.js`
Others are minor fixes like naming convention.

remote-method/USAGE
superkhau: >@superkhau what are your concerns regarding slc loopback:method?

I don't want to explain inconsistent naming in docs/examples and the often overloaded "function/method" terminlogy in discussions. Unless you deem it is safe to assume that all end users will know `method` = `remote-method`. IMO, I'd rather keep `remote-method` = `Remote method` everywhere. No explaining differences etc.

I'm not strongly opinionated here, so I'm fine with avoiding maintaining aliases. However I do think from a user perspective, it makes sense to have aliases (ie. `npm install -> npm i`), but I think they won't complain too much about having to a type an extra few letters. ;)

bajtos: > I don't want to explain inconsistent naming in docs/examples and the often overloaded "function/method" terminlogy in discussions. Unless you deem it is safe to assume that all end users will know method = remote-method. IMO, I'd rather keep remote-method = Remote method everywhere. No explaining differences etc.

I see, that makes sense. I guess I have lived too long inside the LoopBack bubble and cannot see this difficulties anymore.

Let's stick to `loopback:remote-method` then.

test/remote-method.test.js
bajtos: > I prefer to apply it and readJsonSync together after this pr

Sounds good!

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
+'use strict';
+var yeoman = require('yeoman-generator');
+var chalk = require('chalk');
+
+var actions = require('../lib/actions');
+var helpers = require('../lib/helpers');
+var validateName = helpers.validateName;
+var typeChoices =helpers.getTypeChoices;
```
bajtos: style: add a space after the assignment operator. (looks like a typo)

```js
var typeChoices = helpers.getTypeChoices;
//               ^
```

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+  // as loopback-workspace is editing (modifying) files when
+  // saving project changes.
+  constructor: function() {
+    yeoman.generators.Base.apply(this, arguments);
+
+    this.argument('modelName', {
+      desc: 'Name of the model',
+      required: false,
+      type: String
+    });
```
bajtos: How about accepting `methodName` as an argument too? I'd prefer to either allow both names as arguments, or not implement any arguments at all.

I'm guessing there may be a reason why it's easier to not support `methodName`. If that's the case, then could you please @jannyHou explain it here?

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+
+  findModelDefinition: function() {
+    this.modelDefinition = this.projectModels.filter(function(m) {
+        return m.name === this.modelName;
+    }.bind(this))[0];
+
+    if (!this.modelDefinition) {
+        var msg = 'Model not found: ' + this.modelName;
+        this.log(chalk.red(msg));
+        this.async()(new Error(msg));
```
bajtos: I think it's ok to simply throw the error instead of calling `this.async()`. Could you please check?

```js
throw new Erorr(msg);
```

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+    this.prompt(prompts, function(answers) {
+      this.methodName = answers.methodName;
+      this.isStatic = answers.isStatic;
+      this.description = answers.description;
+      done();
+    }.bind(this));
+  },
+
+  delimEndpoints: function() {
+    this.log('Let\'s add some Endpoints now.\n');
```
bajtos: Please add `HTTP` somewhere, e.g. `Let\'s add some HTTP Endpoints now.\n`

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  askForEndpoints: function() {
+    var done = this.async();
+
+    var prompts = [
+      {
+        name: 'httpPath',
+        message: 'Enter path for endpoints',
```
bajtos: `Enter the HTTP path of this endpoint`.

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+      done();
+    }.bind(this));
+  },
+
+  delimEndpoints: function() {
+    this.log('Let\'s add some Endpoints now.\n');
+  },
+
+  askForEndpoints: function() {
+    var done = this.async();
```
bajtos: Add info for the user how to break the loop. See e.g. the model generator ([source](https://github.com/strongloop/generator-loopback/blob/99c66a66d4d1e1179749d792ac3c298fd825636c/model/index.js#L170)).

```js
var done = this.async();
this.log('Enter an empty path when done.');
```

remote-method/index.js
bajtos: Edit: on the second thought, I am not sure if "HTTP path" makes much sense.

Let's use `Enter the path of this endpoint:`

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  askForAccepts: function() {
+    var done = this.async();
+
+    var prompts = [
+      {
+        name: 'acceptsArg',
+        message: 'Enter name for accept argument:',
+        required:true
```
bajtos: style: add a space after colon.

```js
required: true
```

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+  },
+
+  askForReturns: function() {
+    var done = this.async();
+
+    var prompts = [
+      {
+        name: 'returnsArg',
+        message: 'Enter name for return argument:',
+        required:true
```
bajtos: style: add a space after colon.

```js
required: true
```

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+      var subprompts = [
+        {
+          name: 'returnsType',
+          message: 'The type of return argument:',
+          type: 'list',
+          choices: typeChoices
+        },
+        {
+          name: 'returnsRoot',
+          message: 'Is it root',
```
bajtos: `Set the response body to the value of this argument (is it root)?`

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+      }.bind(this));
+    }.bind(this));
+  },
+
+  delimAccepts: function() {
+    this.log('Let\'s add some Accept Arguments now.\n');
+  },
+
+  askForAccepts: function() {
+    var done = this.async();
```
bajtos: Add info for the user how to break the loop. See e.g. the model generator ([source](https://github.com/strongloop/generator-loopback/blob/99c66a66d4d1e1179749d792ac3c298fd825636c/model/index.js#L170)).

```js
var done = this.async();
this.log('Enter an empty name when done.');
```

remote-method/index.js
```diff
@@ -0,0 +1,285 @@
...
+      }.bind(this));
+    }.bind(this));
+  },
+
+  delimReturns: function() {
+    this.log('Let\'s add some Return Arguments now.\n');
+  },
+
+  askForReturns: function() {
+    var done = this.async();
```
bajtos: Add info for the user how to break the loop. See e.g. the model generator ([source](https://github.com/strongloop/generator-loopback/blob/99c66a66d4d1e1179749d792ac3c298fd825636c/model/index.js#L170)).

```js
var done = this.async();
this.log('Enter an empty name when done.');
```

lib/helpers.js
```diff
@@ -131,3 +132,14 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+/**
+ * Get the available type choices that should be used for arguments
+ * from workspace.
+ */
+exports.getTypeChoices = function() {
+    var typeChoices = ModelProperty.availableTypes.concat({
+      name: '(other)',
+      value: null
+    });
+    return typeChoices;
```
bajtos: indentation: please use 2 spaces, not 4.

bajtos:
@jannyHou thank you for the update, I think you are almost done here :)

Besides the minor style and UX related comments, there is this question of generator arguments - see https://github.com/strongloop/generator-loopback/pull/124#discussion-diff-46545163

>  I think the function is alright but I am not so satisfied with my code, I would make it concise tomorrow if possible:)

Sounds like a good plan to me :)

remote-method/index.js
jannyHou: No problem I will add it back, yesterday I misunderstood your feedback, sorry for that. It's not a big issue :)

remote-method/index.js
jannyHou: hmm...actually I use `this.async()` to keep it consistent with other files, eg:
https://github.com/strongloop/generator-loopback/blob/master/relation/index.js#L54
https://github.com/strongloop/generator-loopback/blob/master/property/index.js#L56
And I tried `throw new Error(msg)` it has the same performance, so that means somewhere in yeoman catches that error. I will check yeoman doc to confirm that. 
And could you please explain it more like why `throw` is simpler than `this.async` ? Both of them work me but I am a little confused with their difference (in terms of speed, etc), thank you so much.

bajtos: > hmm...actually I use this.async() to keep it consistent with other files, eg:
https://github.com/strongloop/generator-loopback/blob/master/relation/index.js#L54
https://github.com/strongloop/generator-loopback/blob/master/property/index.js#L56

Ah, I didn't realise that, let's keep your current solution then! Consistency is important.

> And I tried throw new Error(msg) it has the same performance, so that means somewhere in yeoman catches that error. I will check yeoman doc to confirm that. 
> And could you please explain it more like why throw is simpler than this.async ? Both of them work me but I am a little confused with their difference (in terms of speed, etc), thank you so much.

I guess this is subjective. As I read the code, it's switching the action to an async one, only to immediately and synchronously end the action with an error. When reading code like this, it forces me to stop and re-read the line again to understand what's actually happening there, it's not immediately obvious on the first sigh (to me).

OTOH, `try new Error(msg)` is very clear (to me).

The micro-performance of `this.async(err)` may be better than `throw err`, because IIRC v8 does not optimise methods containing exception handling code (although I am not sure if it applies to `try/catch` blocks only or whether `throw` is affected too). However, in this particular case, the impact on runtime performance is negligible.

Anyways, as I wrote at the top, let's stay consistent with other parts of this codebase.

remote-method/USAGE
superkhau: >I guess I have lived too long inside the LoopBack bubble and cannot see these difficulties anymore.

:laughing:

remote-method/index.js
jannyHou: ok I keep it as before. And thank you for the detailed explanation!

jannyHou:
hey @bajtos I added http mapping and a few input validations in this commit, and added `method-name` back as the second argument.
So now we can run the command as `slc loopback:remote-method modelName myMethodName`.

remote-method/index.js
```diff
@@ -0,0 +1,315 @@
...
+          message: 'Description of accept argument:',
+          required: true
+        },
+        {
+          name: 'httpSource',
+          message: 'Mapping to http source:',
+          default: 'body',
+          type: 'list',
+          choices: [{
+            name: '(default) No mapping',
```
bajtos: This is not correct. When there is no http mapping specified, strong-remoting will search all places for the argument, see the [source](https://github.com/strongloop/strong-remoting/blob/32efdc433f8bd12f1ecc7872c855216b99c0911e/lib/http-context.js#L246-L258).

That's why I'm proposing to use an option name `(default)`.

remote-method/index.js
```diff
@@ -0,0 +1,315 @@
...
+            'http': {source: answers.httpSource}
+          });
+        } else {
+          accepts.push({
+            'arg': argName,
+            'type': answers.acceptsType,
+            'required': answers.acceptsRequired,
+            'description': answers.acceptsDes
+          });
+        }
```
bajtos: There is no need to quote keys and please avoid code duplication.

A possible solution (feel free to use your own):

```js
var entry = {
  arg: argName,
  type: answers.acceptsType,
  required: answers.acceptsRequired,
  description: answers.acceptsDes,
};

if (answers.httpSource) {
  entry.http = { source: answers.httpSource };
}

accepts.push(entry);
```

bajtos:
@jannyHou thank you for the update. I left two minor comments for the new code. There are few older comments not resolved, PTAL at them too.

> For this feature: 'a dummy method implementation', I could add some tips print on screen but I prefer to implement adding dummy method somewhere then call the api in generator to update it in modelName.js, just like how I call modelDefinition.method.create(). I am not so familiar with the architecture and considering should that be implemented in datasource-juggler or workspace?
Thanks a lot.

Editing arbitrary javascript code is hard, I don't think it's feasible to implement an API like `modelDefinition.methods` for working with the code file.

I was thinking about this recently, and I think the best solution in the terms of effort needed vs. value added is to print an empty method implementation as the last step (action) of the generator. Ideally the printed code should take into account all data the user has entered - model & method names, whether it's a static or a prototype method, "accepts" arguments, "returns" arguments.

For example, let's take the following method data:

```js
{
modelName: 'Machine',
methodName: 'statistics',
isStatic: false,
description: 'Get statistical analysis of machine\'s reliability.',
accepts: [
  { arg: 'from', type: 'Date', description: 'Start date' },
  { arg: 'to', type: 'Date', description: 'End date' },
],
returns: [
  { arg: 'reliability', type: 'number' },
  { arg: 'incidents', type: ['object'] }
]
}
```

The generator should print a code similar to this:

```js
/**
  * Get statistical analysis of machine's reliability
  *
  * @param {Date} from Start date
  * @param {Date} to End date
  * @param {Function(err, Number, Array)} callback
  */
Machine.prototype.statistics = function(from, to, callback) {
  var reliability, incidents;
  // TODO
  callback(null, reliability, incidents);
};
```

remote-method/index.js
bajtos: FWIW, this has been resolved.

remote-method/index.js
bajtos: :+1:

jannyHou:
hey, @bajtos I printed dummy method in the format you suggested:
```
modelName.prototype.methodName = function(accept1, accept2, callback) {
  var return1, return2;
  // TODO
  callback(null, return1, return2);
};
```
`prototype` would print if `isStatic === false`
And if there is no return argument provided,the variable declaration line would not display.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+/**
+ * Print a sample dummy method implemented in `modelName.js` file
+ * @param {String} modelName Name of model
+ * @param {String} methodName Name of remote method
+ * @param {Boolean} isStatic Method defined on constructor with value true
+ *  while defined on prototype with value false
+ * @param {Array} accepts An array of accept arguments
+ * @param {Array} returns An array of return arguments
+ */
+exports.printDummyMethod = function(
```
bajtos: @jannyHou Do you envision to call this method from other places than the remote-method generator? If not then it would be better to keep the implementation in `remote-method/index.js`.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+) {
+  String.prototype.splice = function(idx, str) {
+    var len = this.length;
+    return this.slice(0, len-idx) + str + this.slice(len-idx);
+  };
+
+  var comma = ', ';
+  var acceptArg = '';
+  var returnArg = '';
+  var functionDef = '= funtion (callback) {' + '\n' +
```
bajtos: Please simplify to

```js
var functionDef = '= funtion(callback) {\n' +
```

Also we usually don't add a space between `function` and the opening `(`.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+  '  //TODO \n' +
+  '  callback(null' + '); \n'+
```
bajtos: Trailing space in both lines above, please remove. I mean the space before `\n` in strings.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+  '}';
```
bajtos: Add `;` after the assignment.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+  var prototype = '.prototype';
+
+  functionDef = methodName.concat(functionDef);
+  functionDef = '.'.concat(functionDef);
+  if(!isStatic) functionDef = prototype.concat(functionDef);
+  functionDef = modelName.concat(functionDef);
```
bajtos: Could you please simplify and rewrite using `+` operator instead of `concat`?

E.g.

```js
var functionDef = modelName + (isStatic ? '.' : '.prototype.') + methodName + ' = function(callback) {\n' +
  // etc. +
  '};\n';
```

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+
+  if(accepts.length>0) {
+    accepts.forEach(function(accept) {
+      acceptArg = acceptArg.concat(accept.arg);
+      acceptArg = acceptArg.concat(comma);
+    });
```
bajtos: Please use `Array.prototype.map` and `Array.prototype.join` instead.

```js
var acceptArts = acceps.map(function(a) { return a.arg; }).join(', ');
```

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+    functionDef = functionDef.splice(42, acceptArg);
```
bajtos: Hardcoding `42` is rather brittle. Please use string-based replace instead.

```js
var acceptArts = acceps.map(function(a) { return a.arg; }).concat(['callback']).join(', ');
functionDef = functionDef.replace(/(callback)/, '(' + acceptArg + ')');
```

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+    var lastReturn = returns.pop();
+    functionDef = functionDef.splice(30, varDeclare);
+    returns.forEach(function(_return) {
+      returnArg = returnArg.concat(_return.arg);
+      returnArg = returnArg.concat(comma);
+    });
+    returnArg = returnArg.concat(lastReturn.arg);
+    functionDef = functionDef.splice(32, returnArg);
+    returnArg = comma.concat(returnArg);
+    functionDef = functionDef.splice(5, returnArg);
```
bajtos: Please apply similar changes here too. Use `.map` and `.join` to build the list of callback arguments, use regex instead of hard-coded string indices.

lib/helpers.js
```diff
@@ -131,3 +132,77 @@ exports.getBaseModelForDataSourceName = function(name, allDataSources) {
...
+  }
+
+  console.log('This is a sample dummy method. Please implement it in ' +
+    modelName +
+    '.js:');
```
bajtos: - let's rephrase this sentence a bit:
    ```
    We have added strong-remoting metadata for your new method to %s.
    Now it's up to you to provide implementation in %s.
    Here is a sample code to get you started:

    MyModel.mymethod = function(callback) {
      // TODO
      callback(null);
    };
    ```

- use full relative path, not just a file name, i.e. `common/mymodel.json` and `common/mymodel.js`, where `common` is the *facet* name (you can get it from a property of `ModelDefinition`).
- use Yeoman's logging methods instead of `console.log`, consider adding some colours
- add an empty line to visually delimit the text from the method body

bajtos:
@jannyHou thank you for the update, I left few comments how to improve the solution.

Also please add sample jsdoc comment per https://github.com/strongloop/generator-loopback/pull/124#issuecomment-161933546

lib/helpers.js
bajtos: (The first `%s` in the example above is the model json file, the second `%s` is the model js file.)

lib/helpers.js
superkhau: +1

jannyHou:
hey @bajtos , I applied your feedback and uploaded a sample screenshot on slack.
I also moved the function into `index.js`, that's more easy to log everything.
hmm...any ideas on chalk? I searched through the repository but didn't get many clues about the proper usage of it :) I used yellow to chalk the file that need edit and green to chalk all of the arguments(both return and accept).

remote-method/index.js
```diff
@@ -0,0 +1,379 @@
...
+
+    this.modelDefinition.methods.create(def, function(err) {
+      helpers.reportValidationError(err, this.log);
+      return done(err);
+    }.bind(this));
+  },
+
+  buildSampleRemoteMethodSource: function() {
+    var jsFilePath = this.modelDefinition.facetName + '/' +
+    this.modelName + '.js';
```
bajtos: I think it's better to use `path.join`, so that you get `common\my-model.json` on Windows. Thoughts?

remote-method/index.js
```diff
@@ -0,0 +1,379 @@
...
+      http: https
+    };
+
+    this.modelDefinition.methods.create(def, function(err) {
+      helpers.reportValidationError(err, this.log);
+      return done(err);
+    }.bind(this));
+  },
+
+  buildSampleRemoteMethodSource: function() {
```
bajtos: This method is actually printing the source, so `printSampleRemoteMethodSource` would be a better name.

remote-method/index.js
bajtos: Actually the file name is not just a model name with `.js`/`.json`, you should use `ModelDefinition.getPath(this.modelDefinition.facetName, this.modelDefinition)` to get the path of the model JSON file and `this.modelDefinition.getScriptPath()` to get the path of the `.js` file.

See https://github.com/strongloop/loopback-workspace/blob/73c2fd0e624a64c6e3beee9203caf867d37825b2/common/models/model-definition.js#L91-L106 and https://github.com/strongloop/loopback-workspace/blob/73c2fd0e624a64c6e3beee9203caf867d37825b2/common/models/model-definition.js#L195-L203 for the implementation of those methods.

bajtos:
@jannyHou 

1) The body of is `buildSampleRemoteMethodSource` rather long and difficult to visually navigate. It's a good practice to add an empty line as a delimiter between blocks of code, e.g.

```js
  jsonFilePath = chalk.yellow(jsonFilePath);

  // Describe what's next step
```

However, if you feel like adding a comment to introduce a block of code, then it's usually a sign that the block should be extracted into a standalone function, I get back to this below.

2) The inline `\n` characters make it a bit difficult to read the code, could you use `join('\n')` instead, as in this example?

```js
[
  'line1',
  'line2',
  'line3 ' + myVar + ' here'
].join('\n')
```

3) Instead of string concatenation via `+`, using `util.format` may make the code easier to read:

```js
util.format('line3 %s here', myVar)
```

4) Now back to extracting smaller functions. That would also help the function body (implementation) to stay at a single level of abstraction. So for example, `buildSampleRemoteMethodSource` should contain high-level instructions like "add what's next step, add jsdoc, add dummy function body, print the result" and each individual step should call a method that would deal with low-level implementation details of array-processing and string manipulation.

A mockup implementation to illustrate my point:

```js
printSampleRemoteMethodSource: function() {
  var text = [
    buildIntroduction(this.modelDefinition),
    buildJsdoc(this),
    buildMethodSource(this)
  ].join('\n');
  console.log(text);
}

function buildMethodSource(def) {
  var ref = def.modelName + (def.isStatic ? '.' : '.prototype.') + def.methodName
  var sourceCode = [
    ref + ' = function(callback) {',
    '  // TODO',
    '  callback(%s);',
    '};'
  ].join('\n');

  var inputArgs = buildInputArgs(def.accepts);
  sourceCode = sourceCode.replace(/*...*/, inputArgs);

  var varDeclarations = buildVarDeclarations(def.returns);
  // string replace

  var callbackArgs = buildCallbackArgs(def.returns);
  // string replace

  return sourceCode;
}
```

Perhaps we could do this refactoring together later today?

remote-method/index.js
```diff
@@ -0,0 +1,379 @@
...
+    // Describe what's next step
+    var tip = 'We have added strong-remoting metadata for your new method to ' +
+      jsonFilePath + '\nNow it\'s up to you to provide implementation in ' +
+      jsFilePath + '\n' +
+      'Here is a sample code to get you started:\n';
+    // Describe the format of jsdoc
+    var jsdoc ='/**\n * ' + this.description + '\n*/\n';
+    // Describe what dummy function looks like
+    var functionDef = ' = funtion(callback) {\n' +
+      '  //TODO\n' +
```
bajtos: Nitpick: we add a space after `//`, i.e.`// TODO`

jannyHou:
hey @bajtos thank you so much for correcting my code convention problems. According to your feedback, I mainly did four changes to refactor `printSampleRemoteMethodSource`:
1. Break it into small functions
2. Replace `+` with `util.format()` whenever `+` is used more than once
3. Use `.join('\n')` to connect multiple lines. 
4. Use modelDefinition apis to get file path.
If there is anything to be improved please tell me, thanks a lot!

bajtos:
Hi @jannyHou, the code building jsdoc comments looks much better now, I find it much easier to read and understand.

I left three more comments to address, see above.

jannyHou:
hey @bajtos I hope after my change `printSampleRemoteMethodSource` looks more readable now.
I rebased the branch to make history cleaner and also added description on it.

bajtos:
@jannyHou I made few final cleaning myself in 3d934be to speed things up.

bajtos:
Landed :tada: :clap:

PR #127, opened 2015-12-05
fix assumptions about dependency locations

rmg:
These assumptions break under npm@3 due to the flattening and
deduplication that is now done.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -584,7 +584,7 @@ module.exports = yeoman.generators.Base.extend({
 
     // Hack: create a clone of the environment because we don't want to share
     // the runLoop. Based on yeoman-generator/lib/actions/invoke
-    var YeomanEnv = require('yeoman-generator/node_modules/yeoman-environment');
+    var YeomanEnv = this.env.constructor;
     var env = YeomanEnv.util.duplicateEnv(this.env);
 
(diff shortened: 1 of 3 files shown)


rmg:
@raymondfeng @bajtos @jannyHou this change is pretty minor and it allows the tests to pass with npm@3, which is what comes with node-v5 and is used.

Since this module is built as part of PRs for loopback and other modules and all the spam on Slack about failed builds, I'm just going to merge this as is.

example/index.js
```diff
@@ -584,7 +584,7 @@ module.exports = yeoman.generators.Base.extend({
 
     // Hack: create a clone of the environment because we don't want to share
     // the runLoop. Based on yeoman-generator/lib/actions/invoke
-    var YeomanEnv = require('yeoman-generator/node_modules/yeoman-environment');
+    var YeomanEnv = this.env.constructor;
```
bajtos: Lovely! This is a very elegant solution, I wish I came up with it myself :)

bajtos:
:+1:

example/index.js
rmg: I'm really glad you approve, I was worried I'd be asked to change it and end up apologizing and reverting it! :smiley:

jannyHou:
@rmg :+1:  Thank you so much!

Issue #103, opened 2015-05-14
`slc loopback relation` should not offer build-in models for "from"


transedward:
as mentioned in #35

https://github.com/transedward/loopback-sandbox(the same box I tried reduplicating with my question)

![screen shot 2015-05-14 at 11 07 25 pm](https://cloud.githubusercontent.com/assets/7189930/7634531/133c4236-fa8e-11e4-9c8e-53de1ad6c642.png)

I used `slc loopback:relation` with two built-in models: AccessToken and User as you can see
but After finishing, there is nothing would be shown to proven there are relations between these two models

As you can see, `git status` show there is nothing.

if we use `slc loopback:model` to build a new model then it would show on common/models
then after `slc loopback:relation`, we can see the relation on relation property

what is the best practice of this problem, create every model u gonna use? but then what is built-in for?
or is there some 'magic' after we relation built-in models? then what if we wanna change afterwards? how can we know it's actually effecting?

transedward:
seems like the same problem was discussed on strongloop/loopback#397
hope to see the new solution coming out
but currently, relation still can be defined on generator but can't be seen on json file seems pretty weird and bug-prone.
I don't think it's a good idea.
Probably prohibit this approach and prompt user to do it in built-model.js is better
@bajtos

bajtos:
+1 for modifying the generator to prevent users from creating a relation on a model that cannot be changed (e.g. a built-in model).

jannyHou:
In pr https://github.com/strongloop/generator-loopback/pull/133,
I removed readonly models from choice in `acl`, `property` and 'relation' generators.
Since pr is merged, I am closing it here. thank you!

PR #133, opened 2016-01-06
Prevent readonly models in relation generator

jannyHou:
This pr is created for preventing users from creating a relation on a read-only model

Connect to strongloop/generator-loopback#103

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -95,6 +95,9 @@ actions.saveProject = function() {
  * Load all models of the current project.
  * `this.projectModels` will contain an array of all models (Array.<Model>)
  * `this.modelNames` will contain an array of names (Array.<string>)
+ * `this.editableModels` will contain an array of models without read-only
+ * models (Array.<Model>)
+ * `this.editableModelNames` will contain an array of names (Array.<string>)
  */
 actions.loadModels = function() {
   var done = this.async();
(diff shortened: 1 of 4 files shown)


jannyHou:
hey @bajtos I am trying to solve the problem in https://github.com/strongloop/generator-loopback/issues/103 by providing readonly Models in prompt choices.
Do you have any ideas on this solution? Thanks!

And if it works for relation generator, I think the same choices should apply to acl and model generators.
I haven't check other generators carefully and this is just a thought, because my strongloop suddenly crushed(I got "Error loading module loopback" with `slc loopback`), so sorry for that! I am trying my best to fix my local strongloop.

lib/actions.js
```diff
@@ -104,6 +107,12 @@ actions.loadModels = function() {
     this.modelNames = results.map(function(m) {
       return m.name;
     });
+    this.editEnableModels = results.filter(function(result) {
```
bajtos: I find it difficult to understand the name `editEnableModel`, how about `editableModels` or `writeableModels`? Similarly for the list of model names below.

bajtos:
Hi @jannyHou, this is an elegant solution! 

I wish we had infrastructure in place that would allow us to write an automatic test for changes like these. Never mind, let's leave that out of scope of this PR.

> I think the same choices should apply to acl and model generators.

ACL generator - yes, that makes sense.
Mode generator is creating a new generator, which is by definition editable/writable, right?

Would you like to fix the ACL generator as part of this PR, since the change should be pretty small?

jannyHou:
hey @bajtos ,thank you! And sure I will update acl generator as well.

lib/actions.js
jannyHou: sure `editable` sounds perfect!

bajtos:
Great stuff, the patch LGTM now.

@jannyHou Please squash the commits to a single one and edit the description to mention all three affected generators. Make sure the first line (commit summary) fits into 50 characters, see the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). You can land the patch yourself then, no need for any further reviews.

PR #134, opened 2016-01-07
Fix ci timeout error.

0candy:
43 passing (51s)
  1 failing

  1) loopback generator help "before each" hook: createSandbox for "print help message with yo by default":
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


superkhau:
LGTM

PR #136, opened 2016-01-13
Prevent constructor to be property name

jannyHou:
Connect to strongloop-internal/scrum-loopback#642

Hey @bajtos this pr is created for preventing users naming a property as `constructor` in property, relation and remote method generators.
I use an array to contain reserved keywords in case we have more in the future, but now it only has `constructor`. 
Could you review it? Thanks!

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -5,6 +5,8 @@ var yeoman = require('yeoman-generator');
 var Workspace = require('loopback-workspace').models.Workspace;
 var ModelProperty = require('loopback-workspace').models.ModelProperty;
 
+var RESERVED_PROPERTY_NAMES = ['constructor'];
+
 exports.reportValidationError = function(err, logFn) {
   if (!err || err.name !== 'ValidationError') {
     return;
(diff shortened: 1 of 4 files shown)


bajtos:
Hi Janny, please refactor the code and create a new method `validatePropertyName`, this method should call `validateName` and then check reserved keywords. Then you can use this methods instead of doing two checks `validateName && checkReserveKeywords`.

lib/helpers.js
```diff
@@ -44,6 +44,22 @@ exports.validateAppName = function (name) {
 };
 
 /**
+ * Check reserved keywords for name of properties and relations
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.checkReserveWords = function(name) {
+  var reserveWords = ['constructor'];
```
bajtos: Looks like a typo, I think this should be `reservedWords`. Since it's a constant, I am proposing to move it to top-level and use an upper-case name:

```
var RESERVED_PROPERTY_NAMES = ['constructor'];
```

lib/helpers.js
```diff
@@ -44,6 +44,22 @@ exports.validateAppName = function (name) {
...
+  var conflict = reserveWords.filter(function(rw) {
+    return rw===name;
+  });
+  if(conflict.length>0) {
```
bajtos: I think a more common solution is to use `Array.prototype.indexOf`: 

```js
if (reserveWords.indexOf(name) !== -1) // then report error
```

jannyHou:
Thanks you so much @bajtos I applied your feedback.

lib/helpers.js
```diff
@@ -44,6 +46,21 @@ exports.validateAppName = function (name) {
 };
 
 /**
+ * Validate property name
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.checkPropertyName = function(name) {
+  if (exports.validateName(name) !== true) {
+    return exports.validateName(name);
```
bajtos: Please cache the result of `validateName`.

```js
var result = exports.validateName(name);
if (result !== true) return result;
```

bajtos:
@jannyHou Nice! One more comment to address, the patch LGTM otherwise. Feel free to land it without waiting for further review.

lib/helpers.js
jannyHou: thanks! I applied it.

PR #138, opened 2016-01-30
Import `generator#invoke()` directly

### LINKED ISSUE #116: deprecated Yeoman generator#invoke called from loopback:model on property creation
```
Enter an empty property name when done.
? Property name: name
(!) generator#invoke() is deprecated. Use generator#composeWith() - see http://yeoman.io/authoring/composability.html
   invoke   loopback:property
? Property type: (Use arrow keys)
❯ string
  number
  boolean
  object
  array
  date
  buffer
```
not clear if #109 fixes this, I don't see references to `invoke` in the patch.

bajtos:
Import the implementation of `generator#invoke` directly from `generator-yeoman/actions/invoke` to prevent a deprecation warning.

Fix #116

@jannyHou could you please verify that my solution proposed here works for you too?

/to @raymondfeng please review

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -30,7 +30,7 @@ module.exports = yeoman.generators.Base.extend({
     // A workaround to get rid of deprecation notice
     //   "generator#invoke() is deprecated. Use generator#composeWith()"
     // See https://github.com/strongloop/generator-loopback/issues/116
-    this.invoke = require('yeoman-generator/actions/invoke');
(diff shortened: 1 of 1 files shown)


raymondfeng:
Nice trick for now. LGTM

jannyHou:
@slnode test please

jannyHou:
Thank you @bajtos ! This is a very smart solution :)
I correct the path to require function `invoke` and it works now.

PR #140, opened 2016-01-31
Prompt for Default Value on Property creation

cmoore4:
adds a prompt for a default value when a new property is created on a model.  Currently, it supports the boolean values of true and false, the special cases of "guid", "uuid", "now" on date/datetime, as well as strings.  

I'm not super in love with this nested if/else structure, but since everything comes back from the command line as a string which we have to parse (based on the property's type), I'm not immediately seeing a better way to do this.  Additionally "guid", "uuid", and "now" are special cases as they set the property "defaultFn" as opposed to just "default".  It's not elegant, but it does work.

### DIFF
--- a/property/index.js
+++ b/property/index.js
@@ -5,7 +5,6 @@ var chalk = require('chalk');
 var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 var validateName = helpers.validateName;
-var checkPropertyName = helpers.checkPropertyName;
 var typeChoices = helpers.getTypeChoices();
 
 module.exports = yeoman.generators.Base.extend({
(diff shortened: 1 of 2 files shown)


slnode:

raymondfeng:
@cmoore4 Nice work. Would you like to add a test case into https://github.com/strongloop/generator-loopback/blob/master/test/property.test.js?

slnode:

slnode:

cmoore4:
Sorry for the Close/Re-Open, Travis was having some issues and I needed to re-trigger the build.  Anyway, tests added :)

PR #144, opened 2016-02-10
Helpers to display correct usage message for apic

davidcheung:
This will require apic cli side to set `process.env.SLC_COMMAND` to apic 
Reason being we cannot set `command` directly from `process.env.SLC_COMMAND` is [this use case](https://github.com/strongloop/generator-loopback/blob/master/test/help.test.js#L27)

and `exports.getCommandName` is treating `process.env.SLC_COMMAND` as a boolean flag

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -101,7 +101,15 @@ exports.checkRelationName = function (modelDefinition, name, callback) {
  * @returns {string}
  */
 exports.getCommandName = function() {
-  return process.env.SLC_COMMAND ? 'slc' : 'yo';
+  var command = 'yo';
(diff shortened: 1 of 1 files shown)


davidcheung:
@raymondfeng in #143, you referenced `cmd=='apic'`, I seem to be getting `yo` as cmd, is it dependent on change on the apiconnect-cli-loopback side? or would this be a OS issue(I'm on windows)? 
>eg `process.argv` gives me `apiconnect\bin\cli.js` rather than `/usr/local/bin/apic`

raymondfeng:
LGTM

Issue #139, opened 2016-01-30
Refactor to play well with Yeoman's run loop


bajtos:
A follow-up for https://github.com/strongloop/generator-loopback/issues/116#issuecomment-173134304.

We are using yeoman generators in a way that was supported in pre-v0.17 days, but that is different from the new direction where yeoman is heading nowadays.

Our design is based on assumption that generators work like javascript methods: if we want to compose "loopback:model" with "loopback:property", we can simply `generator#invoke()` the later and the execution of the former would be postponed until the nested generator finishes.

However, that's no longer the case. These days, Yeoman uses a run queue with named groups (see "The Run Loop" in [docs](http://yeoman.io/authoring/running-context.html)). When two generators are composed, their steps/actions are interleaved together via these named groups (e.g. "prompting", "configuring", "writing", "install", etc.). This composition does not allow repeated invocation of a nested generator, as we are doing now.

I think we should rework generator-loopback, refactor most of the prompts & actions so that they don't depend on yeoman features and the yeoman integration becomes just a thin wrapper. As a side effect, it should become easier to write unit-tests for prompts to verify input validation for example.

Mockup for illustration:

``` js
// model generator
generators.Base.extend({
  prompting: {
     promptForModelData: function() { 
       this.modelData = prompts.modelData();
     },
     promptForProperties: function() {
       this.properties = [];
       for (;;) {
         var prop = prompts.propertyData(this.modelData.name);
         if (!prop.name) break;
         this.properties.push(prop);
      }
  },
  writing: {
     createModel: function() {
       scaffold.model(this.modelData);
     },
     createProperties: function() {
       this.properties.forEach(function(prop) {
          scaffold.property(prop);
       });
     }
  }
});
```

bajtos:
This should also allow us to write automated tests that would prevent a situation that happened recently, when we modified prompt validation to return a promise because of a breaking change in yeoman, and then when yeoman published a new release fixing that breaking change, nobody noticed that generator-loopback no longer works, because CI was happily passing.

hacksparrow:
@bajtos why can't we do without Yeoman?

bajtos:
> why can't we do without Yeoman?

At the I opened this issue, we were relying on Yeoman's `yo` command to run our generators (commands). It also seemed like a good idea to play well with Yeoman ecosystem.

Things has changed since then a lot. Among other things, we have https://github.com/strongloop/loopback-cli as the top-level CLI that LoopBack users should use instead of `yo`.

[...]

PR #147, opened 2016-02-17
Use http and verb

jannyHou:
Related to : https://github.com/strongloop/loopback/issues/2070

In [remote-method](https://github.com/strongloop/generator-loopback/blob/master/remote-method/index.js#L160-L161) generator:
change `httpPath` to `path`
change `httpVerb` to `verb`
to make the setting work. `httpPath` and `httpVerb` are not recognized.

Otherwise remote method would use the default setting: `path`=methodname and `verb`=post

### DIFF
--- a/remote-method/index.js
+++ b/remote-method/index.js
@@ -157,8 +157,8 @@ module.exports = yeoman.generators.Base.extend({
 
       this.prompt(subprompts, function(answers) {
         this.http.push({
-          httpPath: httpPath,
-          httpVerb: answers.httpVerb || answers.customHttpVerb
+          path: httpPath,
+          verb: answers.httpVerb || answers.customHttpVerb
         });
         this.log('\nLet\'s add another endpoint.');
         this.askForEndpoints();


jannyHou:
@bajtos Please review it :) Thanks.

---
For the windows test failure in previous pr, there is no failed test case in the error report, error comes from the first running of test, but disappears when jenkins rerun it.
But I got errors from (travis ci)[https://travis-ci.org/strongloop/generator-loopback/builds/109898723], this is not the first time that I see similar timeout error, I will work on it and see if I can improve the tests:
```
1) loopback:acl generator "after each" hook for "skips accessType is the scope is method":

     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

  

  2) loopback:app generator "before each" hook for "creates expected files":

     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

bajtos:
@jannyHou is there any reasonably easy way how to write a unit-test for this?

jannyHou:
@bajtos hmm..how about setup a test like [`rest.test.js`](https://github.com/strongloop/strong-remoting/blob/master/test/rest.test.js#L690-L713) to make sure the remote method works.

Suppose I generate a remote method in `model.json`:
```
 "myremote": {
      "isStatic": true,
      "accepts": [],
      "returns": [
          {
            "arg": "n", 
            "type": "number" 
          }
      ],
      "description": "This is my remote method",
      "http": [
        {
          "path": "/mypass",
          "verb": "delete"
        }
      ]
    },
```

Then test should pass as:
```
it('should allow arguments in the header', function(done) {
      var method = givenSharedStaticMethod(
        function bar(cb) {
          cb(null, 3);
        },
        {
          accepts: [],
          returns: { arg: 'n', type: 'number' },
          http: { verb: 'get', path: '/mypass' }
        }
      );

      request(app).get(method.classUrl)
        .set('Accept', 'application/json')
        .set('Content-Type', 'application/json')
        .send()
        .expect('Content-Type', /json/)
        .expect({ n: 3 }, done);
    });
```
And in this way I guess I need to include `strong-remoting` as devdependant.
To be honest I am not quite familiar with strong-remoting :) Do you think it's a reasonable way to verify it?

jannyHou:
@bajtos I add tests for http setting as I suggested above, could you review it :) thanks!

bajtos:
Hi @jannyHou, I am afraid this new tests does not belong to generator-loopback, a test like that makes sense in the context of strong-remoting.

As far as this pull request is concerned, I think it's enough to extend [this existing test case](https://github.com/strongloop/generator-loopback/blob/03a965f19d6a5c9e673747068b413c43b4481013/test/remote-method.test.js#L33-L57) and make sure that `{ httpPath, httpVerb }` in the prompt is correctly transformed into `{ verb, path }` in model JSON file.

jannyHou:
@bajtos sorry I misunderstood last time lol. 

I tried to extend the mock test, but still could not do it since http prompt is a loop prompt. I try to set flag variable then control the value of [httpPath](https://github.com/strongloop/generator-loopback/blob/03a965f19d6a5c9e673747068b413c43b4481013/test/remote-method.test.js#L40), but since we pass an object to mockPrompt, it seems impossible to control it.

I am working on automation story recently, and theoretically we can test loop prompts with the combination of `tap` and `child_process`, [like this](https://github.ibm.com/apimesh/automation-apiconnect-cli/blob/master/test/apiccli-loopback.js#L218-L242), I tried with `apic loopback:remote-method`, it kinda works but with very confusing bug.

bajtos:
> I tried to extend the mock test, but still could not do it since http prompt is a loop prompt. I try to set flag variable then control the value of httpPath, but since we are passing an object to mockPrompt, it seems impossible to control it.

Oh, I forgot that we have a loop prompt here. In that case, let's land this pull request without unit-tests.

> I am working on automation story recently, and theoretically we can test loop prompts with the combination of tap and child_process, like this, I tried with apic loopback:remote-method, it kinda works but with very confusing bug.

I was experimenting with this approach myself too. In my experience, the biggest issue is that `inquirer` edits the screen quite a lot, which makes it difficult to detect from a script when it's time to enter what input.

> And I am also thinking of the possibility of using it in the future if it could reduce our rely on yeoman-generator. Just a thought, haven't got time to dig into it.

Personally, I'd like to try https://github.com/strongloop/generator-loopback/issues/139 first and see if allows us to write more powerful unit tests.

PR #149, opened 2016-02-24
Add project template prompt, support apic

bajtos:
When creating a new LoopBack project, let the user choose a template they would like to use.

When running from apiconnect CLI, remove "api-server" from the templates list and offer "hello-world" as the default template.

Requires https://github.com/strongloop/loopback-workspace/pull/255

/to @ritch please review
/cc @kraman

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -7,6 +7,7 @@ var Workspace = workspace.models.Workspace;
 
 var fs = require('fs');
 var path = require('path');
+var util = require('util');
 
 var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

app/index.js
```diff
@@ -64,17 +65,23 @@ module.exports = yeoman.generators.Base.extend({
...
-          // name: util.format('%s (%s)', t.name, t.description),
-          // value: t.name
-          name: t,
-          value: t
+          name: util.format('%s (%s)', t.name, t.description),
+          value: t.name
         };
       });
+
+      if (helpers.getCommandName() === 'apic') {
```
ritch: This is not so great. Lets add a TODO or issue to get rid of it.

ritch:
LGTM

app/index.js
bajtos: I think https://github.com/strongloop/generator-loopback/issues/139 will enable us to get rid of this easily. I'll leave a TODO comment.

PR #150, opened 2016-02-25
Add `yo loopback --skip-next-steps` options

bajtos:
Add an option to suppress "next steps" banner, e.g. when running from a wrapper like `apic`.

/to @ritch please review
/cc @kraman

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -28,6 +28,11 @@ module.exports = yeoman.generators.Base.extend({
       desc: 'Do not install npm dependencies.',
       type: Boolean
     });
+
+    this.option('skip-next-steps', {
+      desc: 'Do not print "next steps" info',
+      type: Boolean
+    });
   },
 
   greet: function() {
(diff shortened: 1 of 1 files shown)


ritch:
Will this allow us to remove:

https://github.com/strongloop/generator-loopback/pull/150/files#diff-2018087f584c4398b5c3a23fc0e5f9dbR204

and friends?

bajtos:
> Will this allow us to remove:
>
> https://github.com/strongloop/generator-loopback/pull/150/files#diff-2018087f584c4398b5c3a23fc0e5f9dbR204
>
> and friends?

Yes in "next steps", not [here](https://github.com/strongloop/generator-loopback/blob/54e64102b604b7aa8c213192e0e19dc85cb98536/app/index.js#L79).

The only reason why I didn't remove apic-related `if` in "next steps" was my worry about backwards compatibility.

@ritch Thoughts?

ritch:
LGTM

PR #155, opened 2016-03-07
Remove prompt for datasources when none exist

ritch:
/to @raymondfeng 
/cc @bajtos 

This fixes https://github.ibm.com/apimesh/scrum-platform-foundation/issues/344

```
Right now, if you dont select anything and hit enter, it fails
$ apic loopback => Empty app
$ apic create --type model
? Enter the model name: m5
? Select the data-source to attach m5 to: (Use arrow keys)
(no data-source)

readline.js:924
throw err;
^
```

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -66,6 +66,7 @@ module.exports = yeoman.generators.Base.extend({
 
   askForParameters: function() {
     var done = this.async();
+    var hasDatasources = this.dataSources && this.dataSources.length > 1;
 
     this.displayName = chalk.yellow(this.name);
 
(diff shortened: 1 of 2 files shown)


model/index.js
```diff
@@ -125,8 +118,24 @@ module.exports = yeoman.generators.Base.extend({
       }
     ];
 
+    if (hasDatasources) {
+      prompts.unshift({
```
raymondfeng: I think you can make a prompt conditional with a `when` property.

raymondfeng:
LGTM

PR #158, opened 2016-03-10
Add support for object/array data source settings

raymondfeng:
This PR adds support for object/array data source settings


raymondfeng:
See https://github.com/strongloop/loopback-workspace/pull/263

PR #159, opened 2016-03-14
Fix data source ordier and default for models

raymondfeng:
This PR has two fixes:

1. Make sure `null` data source is the last option of the list
2. Fix how default `db` is tested
3. Default to `db` if it exists, otherwise the first datasource


rmg:
@slnode test please

rmg:
@slnode test please

PR #160, opened 2016-03-15
Add the option to install connector modules on demand

raymondfeng:
/to @ritch

### DIFF
--- a/datasource/index.js
+++ b/datasource/index.js
@@ -9,6 +9,8 @@ var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 var validateName = helpers.validateName;
 var objectValidator = helpers.objectValidator;
+var path = require('path');
+var fs = require('fs');
 
 module.exports = yeoman.generators.Base.extend({
   // NOTE(bajtos)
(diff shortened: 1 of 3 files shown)


datasource/index.js
```diff
@@ -184,6 +191,42 @@ module.exports = yeoman.generators.Base.extend({
     }.bind(this));
   },
 
+  installConnector: function() {
```
ritch: Will this always execute after the selection step? Yeoman has changed slightly since we first used this to guarantee order... wanted to double check.

package.json
```diff
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "pretest": "jshint .",
-    "test": "mocha -R spec"
+    "test": "mocha -R spec --timeout 60000"
```
ritch: I wonder if we can skip the install on CI so we don't need this timeout...

datasource/index.js
raymondfeng: Yes, it will only be executed after the selection step AFAIK.

package.json
raymondfeng: The timeout to run only one test to verify the install. The timeout is just for safety. If `npm i` finishes earlier, the build moves on.

ritch:
Not sure we want to depend on installing in the tests... But this LGTM to merge either way.

package.json
ritch: > The timeout to run only one test to verify the install

....unless you miss a callback in a test. Its not a big deal though.

PR #165, opened 2016-03-24
Exit gracefully when generator errors

kraman:
Currently invalid user input results in an ugly stack trace. This PR changes behavior to skip follow-on steps if one of the previous steps has erred.

```
$ apic  create --type datasource 
? Enter the data-source name: 
? Select the connector for : In-memory db (supported by StrongLoop)
Connector-specific configuration:
? window.localStorage key to use for persistence (browser only): 
? Full path to file for persistence (server only): 
Validation error: invalid DataSourceDefinition
 - name: can't be blank,can't be blank

events.js:141
      throw er; // Unhandled 'error' event
      ^
ValidationError: The `DataSourceDefinition` instance is not valid. Details: `name` can't be blank (value: ""); `name` can't be blank (value: "").
    at /usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback-datasource-juggler/lib/dao.js:317:12
    at ModelConstructor.<anonymous> (/usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback-datasource-juggler/lib/validations.js:487:13)
    at ModelConstructor.next (/usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback-datasource-juggler/lib/hooks.js:75:12)
    at done (/usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback-datasource-juggler/lib/validations.js:484:25)
    at /usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback-datasource-juggler/lib/validations.js:558:7
    at doNTCallback0 (node.js:417:9)
    at process._tickDomainCallback (node.js:387:13)
    at process.<anonymous> (/usr/local/lib/node_modules/apiconnect/node_modules/apiconnect-cli-loopback/node_modules/generator-loopback/node_modules/loopback-workspace/node_modules/loopback/node_modules/continuation-local-storage/node_modules/async-listener/index.js:19:15)
```

PR #169, opened 2016-03-29
Check empty name in validateName and add testcase

jannyHou:
enhancement of pr https://github.com/strongloop/generator-loopback/pull/165#issuecomment-201057249

1. Check empty name in shared helper function validateName()
2. Add test case for rejecting empty name
3. remove duplicate checking of encodeURIComponent(name)

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -65,11 +65,13 @@ exports.checkPropertyName = function(name) {
  * @returns {String|Boolean}
  */
 exports.validateName = function (name) {
-  if (name.match(/[\/@\s\+%:\.]/) ||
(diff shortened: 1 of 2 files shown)


jannyHou:
@bajtos please review:) thanks.
Besides the empty value check, I am not sure is it a duplicate check of encodeURIComponent here:
https://github.com/strongloop/generator-loopback/blob/master/lib/helpers.js#L68-L72
if I understand correctly the current code would never return the second error warning.

bajtos:
LGTM, :shipit:

bajtos:
@jannyHou actually, please squash the commits first, before landing the change :)

PR #172, opened 2016-04-05
Fix how default base model is set

raymondfeng:
This PR makes sure default base model is calculated correctly. We need to select the data source first.


ritch:
LGTM

PR #173, opened 2016-04-07
Allow defining remote method without isStatic flag

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/813

- Update generator-loopback remote-method to make isStatic prompt not required. If not specified, it will default to true if method name does not contain `prototype.` and false otherwise.
- Add new validation for method name to allow for `prototype.` otherwise `.` is still invalid.
- Add unit test cases.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -66,6 +66,24 @@ exports.checkPropertyName = function(name) {
 };
 
 /**
+ * Validate remote method name. Allows `.` in method name for prototype
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.validateRemoteMethodName = function(name) {
+  if (name.match(/\./) && !name.match(/^prototype\.([^.]*)$/)) {
+    return 'Name cannot contain "." characters except' +
+      ' the dot in "prototype." prefix';
+  }
+  var result = exports.validateOptionalName(name, /[\/@\s\+%:]/);
+  if (result !== true) return result;
+  if (RESERVED_PROPERTY_NAMES.indexOf(name) !== -1) {
+    return name + ' is a reserved keyword. Please use another name';
+  }
+  return true;
+};
+
+/**
  * Validate required name for properties, data sources, or connectors
  * Empty name could not pass
  * @param {String} name The user input
@@ -84,9 +102,13 @@ exports.validateRequiredName = function (name) {
  * @param {String} name The user input
  * @returns {String|Boolean}
  */
-exports.validateOptionalName = function (name) {
-  if (name.match(/[\/@\s\+%:\.]/)) {
-    return 'Name cannot contain special characters (/@+%:. ): ' + name;
+exports.validateOptionalName = function (name, unallowedCharacters) {
+  if (!unallowedCharacters) {
+    unallowedCharacters = /[\/@\s\+%:\.]/;
+  }
+  if (name.match(unallowedCharacters)) {
+    return 'Name cannot contain special characters ' + unallowedCharacters +
+     name;
   }
   if (name !== encodeURIComponent(name)) {
     return 'Name cannot contain special characters escaped by ' +
--- a/remote-method/index.js
+++ b/remote-method/index.js
@@ -13,7 +13,7 @@ var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 var validateRequiredName = helpers.validateRequiredName;
 var validateOptionalName = helpers.validateOptionalName;
-var checkPropertyName = helpers.checkPropertyName;
+var validateRemoteMethodName = helpers.validateRemoteMethodName;
 var typeChoices = helpers.getTypeChoices();
 var ModelDefinition = require('loopback-workspace').models.ModelDefinition;
 
@@ -96,14 +96,16 @@ module.exports = yeoman.generators.Base.extend({
         message: 'Enter the remote method name:',
         required: true,
         default: name,
-        validate: checkPropertyName
+        validate: validateRemoteMethodName
       },
       {
         name: 'isStatic',
         message: 'Is Static?',
-        required:true,
+        required: false,
         type: 'confirm',
-        default: true
+        default: function(answers) {
+          return !answers.methodName.match(/^prototype\.(.*)$/);
+        }
       },
       {
         name: 'description',
@@ -111,7 +113,10 @@ module.exports = yeoman.generators.Base.extend({
       }
     ];
     this.prompt(prompts, function(answers) {
-      this.methodName = answers.methodName;
+      var m = answers.methodName.match(/^prototype\.(.*)$/);
+      var isStatic = !m;
+      var baseName = isStatic ? answers.methodName : m[1];
+      this.methodName = baseName;
       this.isStatic = answers.isStatic;
       this.description = answers.description;
       done();
--- a/test/helpers.test.js
+++ b/test/helpers.test.js
@@ -10,6 +10,7 @@ var validateAppName = helpers.validateAppName;
 var validateOptionalName = helpers.validateOptionalName;
 var validateRequiredName = helpers.validateRequiredName;
 var checkRelationName = helpers.checkRelationName;
+var validateRemoteMethodName = helpers.validateRemoteMethodName;
 require('chai').should();
 var expect = require('chai').expect;
 var promise = require('bluebird');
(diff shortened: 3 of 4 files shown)


remote-method/index.js
```diff
@@ -90,14 +90,16 @@ module.exports = yeoman.generators.Base.extend({
...
       {
         name: 'isStatic',
         message: 'Is Static?',
-        required:true,
+        required: false,
         type: 'confirm',
-        default: true
+        default: function(answers) {
+            return !answers.methodName.match(/^prototype\.(.*)$/);
+          }
```
bajtos: Indentation - L101 and L102 are indented by one extra level.

test/helpers.test.js
```diff
@@ -63,6 +64,30 @@ describe('helpers', function() {
...
   });
 
+  // test checkRemoteMethodName()
+  describe('checkRemoteMethodName()', function() {
+    it('should accept good names', function() {
+      testValidationAcceptsValue(checkRemoteMethodName, 'prop');
+      testValidationAcceptsValue(checkRemoteMethodName, 'prop1');
+      testValidationAcceptsValue(checkRemoteMethodName, 'my_prop');
+      testValidationAcceptsValue(checkRemoteMethodName, 'my-prop');
+      testValidationAcceptsValue(checkRemoteMethodName, 'm.prop');
```
bajtos: Please use `prototype.method` to make it clearer that we want to support this particular case.

I think we should not allow `.` in method names except the `/^prototype\./` case, i.e. `m.prop` is not a valid method name.

bajtos:
@0candy could you please describe in pull request description what exactly are you changing here? I am mostly able to get that by reading the code, however other people and especially our users may not.

bajtos:
As I commented in https://github.com/strongloop/loopback-workspace/pull/273, I would like to preserve the current workspace API, i.e. a remote prototype method should still have the name without `prototype.` prefix and `isStatic` flag set to true.

test/remote-method.test.js
```diff
@@ -56,6 +56,32 @@ describe('loopback:remote-method generator', function() {
...
+      httpPath: '',
+      acceptsArg: '',
+      returnsArg: ''
+    });
+
+    methodGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var methods = definition.methods || {};
+      expect(methods).to.have.property('prototype.myRemote');
+      expect(methods['prototype.myRemote']).to.eql({
```
bajtos: This is true when the project is using LoopBack 3.0, i.e. when `common.createDummyProject` creates a LB 3.0 project. However, since 3.0 was not released yet, I think we are using LB 2.0 and thus the entry should use the old convention (name + isStatic). Am I missing something?

test/remote-method.test.js
```diff
@@ -56,6 +56,32 @@ describe('loopback:remote-method generator', function() {
...
+        isStatic: false,
```
bajtos: I find this weird, if your test is assuming static/prototype flag is encoded in the name, then why is there `isStatic` property set too?

bajtos:
The implementation itself looks good to me, however the tests are suspicious. Let's figure out the loopback-workspace part fist and then come back to generator-loopback.

0candy:
The change in generator-loopback was that either methods of defining the remote method should work. Only loopback-workspace distinguishes between LB 2.0 and 3.0.

lib/helpers.js
```diff
@@ -60,6 +60,24 @@ exports.checkPropertyName = function(name) {
 };
 
 /**
+ * Validate remote method name. Allows `.` in method name for prototype
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.checkRemoteMethodName = function(name) {
+  if (name.match(/\./) && !name.match(/^prototype\.(.*)$/)) {
```
bajtos: Is `.` character allowed after `prototype.`? I believe the current code allows that.

I think you need to change the second regexp to `/^prototype\.([^.]*)$/`.

lib/helpers.js
```diff
@@ -60,6 +60,24 @@ exports.checkPropertyName = function(name) {
 };
 
 /**
+ * Validate remote method name. Allows `.` in method name for prototype
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.checkRemoteMethodName = function(name) {
```
bajtos: Please keep the naming consistent and rename this function to `validateRemoteMethodName`.

lib/helpers.js
```diff
@@ -79,15 +97,28 @@ exports.validateRequiredName = function (name) {
...
+  return true;
+};
+
+/**
+ * Validate value and throw error on unallowed characters.
+ * @param {String} name The user input
+ * @param {String} unallowedCharacters Characters that are not allowed.
+ * @returns {String|Boolean}
+ */
+function validateValue(name, unallowedCharacters) {
```
bajtos: The name `validateValue` is not accurate, because it creates impression that any/all values should be validated through this function.

I am proposing to change `validateOptionalName` to accept the new parameter `unallowedCharacters` and call it from `validateRemoteMethodName`. That way we don't need `validateValue` at all.

Thoughts?

lib/helpers.js
bajtos: Oh, I see, we already have `checkPropertyName`. It looks to me that one is an exception, as other method names start with `validate`, so I think it's still better to call the new method `validateRemoteMethodName`.

remote-method/index.js
```diff
@@ -91,14 +91,16 @@ module.exports = yeoman.generators.Base.extend({
...
       },
       {
         name: 'isStatic',
         message: 'Is Static?',
-        required:true,
+        required: false,
         type: 'confirm',
-        default: true
+        default: function(answers) {
+          return !answers.methodName.match(/^prototype\.(.*)$/);
```
bajtos: If we allow `methodName` to start with `prototype.` prefix, then I think we need to remove that prefix before we pass the name to `ModelMethod` constructor. See https://github.com/strongloop/loopback-workspace/pull/273#discussion_r61563760

bajtos:
@0candy could you please add a unit-test to verify that the generator correctly handles the case where the user enters `prototype.instanceMethod` as the method name?

lib/helpers.js
0candy: Doesn't the `\.` after prototype include the `.` character?

bajtos: Here is the test case to add to your unit-tests: 

```js
testValidationRejectsValue(validateRemoteMethodName, 'prototype.dotted.method');
```

lib/helpers.js
```diff
@@ -60,6 +60,24 @@ exports.checkPropertyName = function(name) {
 };
 
 /**
+ * Validate remote method name. Allows `.` in method name for prototype
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.validateRemoteMethodName = function(name) {
+  if (name.match(/\./) && !name.match(/^prototype\.(.*)$/)) {
```
bajtos: > Doesn't the `\.` after prototype include the `.` character?

Here is the test case to add to your unit-tests: 

```js
testValidationRejectsValue(validateRemoteMethodName, 'prototype.dotted.method');
```

remote-method/index.js
bajtos: > If we allow `methodName` to start with `prototype.` prefix, then I think we need to remove that prefix before we pass the name to `ModelMethod` constructor. See https://github.com/strongloop/loopback-workspace/pull/273#discussion_r61563760

@0candy ^^^ thoughts?

0candy: I agree, hence I removed the prefix before passing the name to the constructor.

lib/helpers.js
0candy: ah, sorry. I misunderstood. Thought you said it doesn't allow `prototype.`, but you meant any other `.` after `prototype.`. Fixed.

0candy:
@slnode ok to test

bajtos:
> Here is the test case to add to your unit-tests: 
> ```js
testValidationRejectsValue(validateRemoteMethodName, 'prototype.dotted.method');
```

@0candy I don't see that test case in the unit-tests, is it your intention to leave it out?

File: remote-method/index.js (hunk 8 above)
Comment (bajtos) [inline remote-method/index.js]: Ah, I see, the prefix is removed here: https://github.com/strongloop/generator-loopback/pull/173/files#diff-1162a393eb1e8979663fc397b9fbacbeR113 

✅

**Comment by 0candy** [comment]
I did add it.
https://github.com/strongloop/generator-loopback/pull/173/files#diff-335b7c365ddfd4b8c9f944d8fba9f213R99

File: lib/helpers.js [hunk 10]
```diff
@@ -60,6 +60,24 @@ exports.checkPropertyName = function(name) {
 };
 
 /**
+ * Validate remote method name. Allows `.` in method name for prototype
+ * @param {String} name The user input
+ * @returns {String|Boolean}
+ */
+exports.validateRemoteMethodName = function(name) {
+  if (name.match(/\./) && !name.match(/^prototype\.([^.]*)$/)) {
+    return 'Name cannot contain special characters . ' +
```
bajtos: Nitpick: I think "special characters ." is not grammatically correct. How about the following?

```js
'Name cannot contain "." characters except the dot in "prototype." prefix'
```

bajtos:
One more nitpick, the patch LGTM otherwise. Please squash the commits before landing, no further review is needed.

bajtos:
Oh, and please make sure CI builds are passing, they are all red ATM.

0candy:
CI is all failing on master right now.

0candy:
Yeoman made a breaking change to mockingPrompts. We probably need to update all our unit tests.
https://github.com/yeoman/generator/issues/926

0candy:
Yeoman's latest release has a breaking change in mockPrompts. We probably have to update all our unit tests.
https://github.com/yeoman/generator/issues/926

superkhau:
@0candy Please let me know when you got this working. The fails are also making me cautious of merging https://github.com/strongloop/generator-loopback/pull/180#issuecomment-218345089

PR #174, opened 2016-04-08
Add eslint infrastructure

jannyHou:


package.json
```diff
@@ -14,7 +14,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "pretest": "jshint .",
+    "posttest": "eslint .",
```
superkhau: move this line under "test"

superkhau:
One minor comment, otherwise LGTM

package.json
eddiemonge: why is it being moved to posttest?

jannyHou:
The test cases are broken now so I prefer to finish this pr after it's fixed.

eddiemonge:
closing in favor of #189 which includes this change and also the changes to make eslint happy

package.json
bajtos: When doing red-green-refactor cycle, we don't want to wait for linter every time we run the test suite. Another way of looking at this: first get the implementation right, only then worry about the (coding) style.

PR #175, opened 2016-04-12
Fix accept and return to end with empty input

jannyHou:
Add empty check to `validName` breaks remote-method generator, since in accept and return prompt, it checks validation before checking empty. 
I change the logic in remote-method generator so empty input could pass.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -51,7 +51,7 @@ exports.validateAppName = function (name) {
  * @returns {String|Boolean}
  */
 exports.checkPropertyName = function(name) {
-  var result = exports.validateName(name);
+  var result = exports.validateRequiredName(name);
   if (result !== true) return result;
   if (RESERVED_PROPERTY_NAMES.indexOf(name) !== -1) {
     return name + ' is a reserved keyword. Please use another name';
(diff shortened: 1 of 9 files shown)


jannyHou:
@bajtos Please review :) thanks

bajtos:
@jannyHou @0candy are we sure only remote-method generator is affected?

0candy:
I think anywhere that uses validateName and allows `enter empty name when done` will fail. But I think @jannyHou said some places check for empty and exit before validating.

bajtos:
Makes sense.

We discussed this on Slack with Janny, agreed to refactor the code a bit: rename `validateName` to `validateRequiredName`, add new method `validateOptionalName` and then use these two methods as appropriate. Hopefully it will make it more clear which places allow empty values and which do not.

jannyHou:
@bajtos `validateRequiredName` and `validateOptionalName` are applied. Please review:)

jannyHou:
@slnode test please

lib/helpers.js
```diff
@@ -60,14 +60,27 @@ exports.checkPropertyName = function(name) {
...
  * @returns {String|Boolean}
  */
-exports.validateName = function (name) {
+exports.validateRequiredName = function (name) {
+  var result = exports.validateOptionalName(name);
+  if (result !== true) return result;
   if (!name) {
     return 'Name is required';
   }
+  return true;
```
bajtos: How about simplifying the method by checking whether the name is truthy first?

```js
if (!name) {
  return 'Name is required';
}

return exports.validateOptionalName(name);
```

bajtos:
One more comment (see above). The rest LGTM, no need for further reviews.

bajtos:
> generator-loopback/node=4.x,os=amazon — Failed! (9a81230)

Please check why the build is failing on CI. We should at least get generator-loopback builds green, if not all dependants.

jannyHou:
@slnode test please

lib/helpers.js
jannyHou: agreed:)

Issue #176, opened 2016-04-14
Enhance slc loopback:model generator to display Warning when no datasource available


gunjpan:
Current behavior of `slc loopback:model` generator:
After creating an empty LoopBack project, while creating a new model using `slc loopback:model` or `apic create --type model`, the generator doesn't display following prompt 
```console  
? Select the data-source to attach newModel to: (Use arrow keys)` 
```
until a datasource is defined. Also, it doesn't warn users about it.

This should be enhanced to show a warning message for `ease-of-use`.

richardpringle:
Please add a documentation issue to go along with this and assign it to Rand

0candy:
For doc:
Add a note stating that if you select an empty app it does not come with any ds, not even the in memory one.
So if you go on to run the application, you would not see any of your models as there was no datasource attached.

bajtos:
Published in `generator-loopback@1.22.0`.

PR #177, opened 2016-04-26
Add warning when no datasource available

0candy:
Connect to https://github.com/strongloop/generator-loopback/issues/176

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -45,6 +45,16 @@ module.exports = yeoman.generators.Base.extend({
   loadModels: actions.loadModels,
 
   addNullDataSourceItem: actions.addNullDataSourceItem,
+  
+  checkForDatasource: function() {
+    if (!this.hasDatasources) {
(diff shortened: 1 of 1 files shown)


superkhau:
LGTM

model/index.js
```diff
@@ -45,6 +45,15 @@ module.exports = yeoman.generators.Base.extend({
   loadModels: actions.loadModels,
 
   addNullDataSourceItem: actions.addNullDataSourceItem,
+  
+  checkForDatasource: function() {
+    if (!this.hasDatasources) {
+      var warning = chalk.red('Warning: No data sources available ' +
+        'to attach model.');
```
bajtos: Can we please extend the message and tell the user what are the implications? What should they expect to happen? This way it's hard to tell whether how serious it is that there are not data sources available.

bajtos:
LGTM, :shipit:

PR #180, opened 2016-05-11
Fixes datasource generation with community connector

amuramoto:
Community connectors are not listed in availableConnectors, causing the following when a user chooses 'other' and enters a community connector:

ERROR Cannot read property 'package' of undefined

Fix makes function return when this is the case, allowing generator to proceed creating skeleton record in datasources.json

### DIFF
--- a/datasource/index.js
+++ b/datasource/index.js
@@ -198,6 +198,7 @@ module.exports = yeoman.generators.Base.extend({
 
   installConnector: function() {
     var connector = this.availableConnectors[this.connector];
+    if (!connector) return;
     var pkg = connector.package;
     if (!pkg) return;
 


slnode:

superkhau:
@amuramoto Your PR is failing on Travis, can you check it out?

amuramoto:
@superkhau Tests appear to be failing due to PR #173 which is still under review

superkhau:
@amuramoto Cool, I'll leave this one for @bajtos to review as he is more familiar with this module. It LGTM, but can't really tell with Traavis not passing.

superkhau:
@amuramoto Can you add a unit test to verify your changes?

datasource/index.js
```diff
@@ -198,6 +198,7 @@ module.exports = yeoman.generators.Base.extend({
 
   installConnector: function() {
     var connector = this.availableConnectors[this.connector];
+    if (!connector) return;
```
bajtos: When we don't have any information about the custom connector selected for a new datasource, then I think we should assume the package name is same as the connector name and run `npm install {this.connector}`.

@amuramoto @superkhau @ritch @raymondfeng Thoughts?

bajtos:
@amuramoto thank you for the pull request, this is indeed a bug that needs to be fixed!

I left a comment above, I think we should take a different approach. Also please add a unit-test to verify your change and prevent regressions in the future, see these two existing tests for inspiration: https://github.com/strongloop/generator-loopback/blob/8fbdb0da7ce203d387ee00096f7cfade37a0d3d9/test/datasource.test.js#L45-L81

For posterity, the failing CI build is not waiting for #173, but for fixing our dependencies, per
https://github.com/strongloop/generator-loopback/pull/173#issuecomment-217281272
> Yeoman's latest release has a breaking change in prompts. We probably have to update our code.
yeoman/generator#926 The latest version of yeoman-environment 1.6.1 is breaking the unit test. The promptModule is causing problems.

bajtos:
Closing in favour of #199

PR #181, opened 2016-05-11
Fix unit test with yeoman-test

0candy:
Fix broken unit test from latest dependency update in yeoman-generator. Switch to use yeoman-test instead of yeoman-generator test as its deprecated in the latest release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "glob": "^6.0.1",
     "jshint": "^2.8.0",
     "mocha": "^2.2.5",
-    "strong-cached-install": "^2.0.0"
+    "strong-cached-install": "^2.0.0",
+    "yeoman-test": "^1.4.0"
   }
 }
--- a/test/acl.test.js
+++ b/test/acl.test.js
@@ -11,6 +11,7 @@ var wsModels = require('loopback-workspace').models;
 var SANDBOX =  path.resolve(__dirname, 'sandbox');
 var expect = require('chai').expect;
 var common = require('./common');
+var testhelpers = require('yeoman-test');
 
 describe('loopback:acl generator', function() {
   beforeEach(function createSandbox(done) {
@@ -46,7 +47,7 @@ describe('loopback:acl generator', function() {
 
   it('adds an entry to models.json', function(done) {
     var aclGen = givenAclGenerator();
-    helpers.mockPrompt(aclGen, {
+    testhelpers.mockPrompt(aclGen, {
       model: 'Car',
       scope: 'all',
       accessType: '*',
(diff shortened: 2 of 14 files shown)


superkhau:
LGTM but the tests are still failing.

superkhau:
@slnode test please

bajtos:
Interestingly enough, the tests are passing on Travis CI.

bajtos:
Here is the relevant error:

```
     Error: You don't seem to have a generator with the name loopback:acl installed.
You can see available generators with npm search yeoman-generator and then install them with npm install [name].
To see the 1 registered generators run yo with the `--help` option.
      at Environment.create (node_modules/yeoman-test/node_modules/yeoman-environment/lib/environment.js:296:7)
      at Object.exports.createGenerator (node_modules/yeoman-test/lib/index.js:170:14)
      at Object.createGenerator (test/common.js:19:18)
      at givenAclGenerator (test/acl.test.js:159:22)
      at Context.<anonymous> (test/acl.test.js:123:18)
```

AFAICT, the project is cloned into a directory with a name that does not match the project name (and hence the generator name), I think that's the root cause.

```
[sl-ci-run INFO  0.06 ]: CWD: /mnt/jenkins/workspace/generator-loopback/0fde463e
```

IIRC, @rmg was making some hacks in our CI to fix registration of yeoman generators. Do you remember any details, @rmg?

rmg:
Hm.. I do recall spending a bunch of time investigating this generator problem a while back, but I don't recall what my findings were. Ideally the test helpers would remove this restriction, but I'm not sure if support for that was ever added. I can look into checking the module out into a subdirectory instead of the workspace root.

rmg:
The tests broke because of yeoman/environment@3043b31a41d0bfa385946591cc7e98cf64719877 which replaced an ownProperty `prompt` _object_ with a _method_ on the prototype chain, which causes it to break https://github.com/yeoman/generator/blob/v0.21.1/lib/test/helpers.js#L115

If a dependency on `yeoman-environment@1.5.3` is added, everything works as it did before yeoman-generator changed it's mocking.

Overall, the root of the problem seems to be that yeoman-generator v0.21.1 is really old, but it is the newest version that matches `^0.21.1` because the major version is `0`.

bajtos:
> it is the newest version that matches ^0.21.1 because the major version is 0.

And that's good, because both 0.22.x and 0.23.x introduced backwards-incompatible changes. The root of the problem is that `yeoman/environment` introduced a breaking change in non-major release :(

eddiemonge:
the helper method shouldn't have been renamed. the require should have been changed. as it is now, it leaves an extra unused variable

PR #186, opened 2016-05-18
Refactor callback to use promise

jannyHou:
Connect to strongloop/loopback#2292

### DIFF
--- a/acl/index.js
+++ b/acl/index.js
@@ -13,7 +13,7 @@ var ModelAccessControl = wsModels.ModelAccessControl;
 var actions = require('../lib/actions');
 var helpers = require('../lib/helpers');
 
-module.exports = yeoman.generators.Base.extend({
+module.exports = yeoman.Base.extend({
   // NOTE(bajtos)
   // This generator does not track file changes via yeoman,
   // as loopback-workspace is editing (modifying) files when
--- a/app/index.js
+++ b/app/index.js
@@ -19,9 +19,9 @@ var helpers = require('../lib/helpers');
 var validateAppName = helpers.validateAppName;
 var pkg = require('../package.json');
 
-module.exports = yeoman.generators.Base.extend({
+module.exports = yeoman.Base.extend({
   constructor: function() {
-    yeoman.generators.Base.apply(this, arguments);
+    yeoman.Base.apply(this, arguments);
 
     this.argument('name', {
       desc: 'Name of the application to scaffold.',
--- a/boot-script/index.js
+++ b/boot-script/index.js
@@ -10,9 +10,9 @@ var yeoman = require('yeoman-generator');
 
 var validateRequiredName = helpers.validateRequiredName;
 
-module.exports = yeoman.generators.Base.extend({
+module.exports = yeoman.Base.extend({
   constructor: function() {
-    yeoman.generators.Base.apply(this, arguments);
+    yeoman.Base.apply(this, arguments);
 
     this.argument('name', {
       desc: 'Name of the boot script to create.',
--- a/datasource/index.js
+++ b/datasource/index.js
@@ -17,7 +17,7 @@ var objectValidator = helpers.objectValidator;
 var path = require('path');
 var fs = require('fs');
 
-module.exports = yeoman.generators.Base.extend({
+module.exports = yeoman.Base.extend({
   // NOTE(bajtos)
   // This generator does not track file changes via yeoman,
   // as loopback-workspace is editing (modifying) files when
@@ -26,7 +26,7 @@ module.exports = yeoman.generators.Base.extend({
   loadProject: actions.loadProject,
 
   constructor: function() {
-    yeoman.generators.Base.apply(this, arguments);
+    yeoman.Base.apply(this, arguments);
 
     this.argument('name', {
       desc: 'Name of the data-source to create.',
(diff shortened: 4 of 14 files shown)


jannyHou:
@bajtos This is a quick fix for relation generator.

`yeoman-environment` uses the latest `inquirer`, which depends on a new module `rx` that supports promise only. So we cannot use callback in function [`checkRelationName`](https://github.com/strongloop/generator-loopback/blob/master/relation/index.js#L123)

Please review. Thanks.

jannyHou:
@bajtos I talked to @0candy today, and we agreed that the quickest fix is change the `inquirer` version. 

But if you feel like it's not a recommended solution, I came up with another idea:
yeoman-generator with 0.23.x brings the break-change which destroys all lb-generators but 0.22.x not. And 0.22.x doesn't break the test cases that many(50+ failure when use 0.21.x), Ryan explained the reason: https://github.com/strongloop/generator-loopback/pull/181#issuecomment-219843427

I upgrade dependancy yeoman-generator's version to latest 0.22.x, and with my fix of relation generator, it only has 3 failures. I haven't figured out what's the problem with appName normalization but it does look better than 0.21.x and doesn't need so much work as upgrading to 0.23.x.

CI Failure with the current code in this pr:
```
1) loopback:app generator normalizes the appname with .:
     Error: timeout of 60000ms exceeded. Ensure the done() callback is being called in this test.
  

  2) loopback:app generator normalizes the appname with space:
     Error: timeout of 60000ms exceeded. Ensure the done() callback is being called in this test.
  

  3) loopback:app generator normalizes the appname with @:
     Error: timeout of 60000ms exceeded. Ensure the done() callback is being called in this test.
  
```

Thoughts?

0candy:
We can't downgrade `yeoman-environment` because its set in `yeoman-generator`. The fix I made to get the unit test working is to use `yeoman-test` instead, however that is still breaking on Jenkins but passing on Travis as Ryan did something to make the unit test work on Jenkins. However, that change requires `yeoman-environment` which is what is causing the problems to begin with.

I guess we can't just use `yeoman-test` to fix this problem as the generator is broken for relations as well. I think we need to upgrade `yeoman-generator` to fix both the loopback generator and its test.

jannyHou:
The test for app generator fails because when we build app through `helpters.testDirectory` and `mockPrompt`, see the following code:
```js
helpers.testDirectory(dir, function() {
      console.log('CurrentPath' + path.basename(process.cwd()));
      helpers.mockPrompt(gen, {
        wsTemplate: 'api-server',
        dir: '.',
      });
```
It initializes the `workspace_path` in [`action.js/initWorkspace`](https://github.com/strongloop/generator-loopback/blob/master/lib/actions.js#L62) by:
`this.projectDir = this.destinationRoot();`

But [`destinationRoot()`](https://github.com/yeoman/generator/blob/0644986a01721bd90130e726d66ec20b22ae4535/lib/base.js#L688) always returns path of `sandbox`, because `this.env.cwd` === `path_of_sandbox`, see the following code:

```js

Base.prototype.destinationRoot = function (rootPath) {
  if (_.isString(rootPath)) {
    this._destinationRoot = path.resolve(rootPath);

    if (!pathExists.sync(rootPath)) {
      mkdirp.sync(rootPath);
    }

    process.chdir(rootPath);
    this.env.cwd = rootPath;

    // Reset the storage
    this.config = this._getStorage();
  }

  return this._destinationRoot || this.env.cwd;
};
```

I am confused how does yeoman-generator get the `this.env.cwd`

bajtos:
@jannyHou I don't fully understand the problem you described in your last comment. However, it rang a bell:  IIRC, yeoman checks for presence of `.yo-rc.json` in the current directory and then walks up the directory tree until the file is found. Perhaps it sets destination root to the directory where `.yo-rc.json` is found, is that possible? It's just an idea to check, I may be wrong.

> But destinationRoot() always returns path of sandbox

I think that is intentional, isn't it? We always run the generator in the sandbox directory which is then cleared between each test run. That way the generator has always clean working place.

jannyHou:
@bajtos Yeah it's a little complex to explain, let me try another way:
So now when we run `slc loopback` inside and outside the `strongloop/node_modules/generator-loopback/test/sandbox`, the performance are different.

For example run `slc loobpack` to create an app called `x.x` outside it, it works fine, `this.env.cwd` returns `DIR_PATH/x.x/x-x` .
But run `slc loobpack` inside sandbox folder, it should return `sandbox/x.x/x-x` but it actually returns `sandbox`.

bajtos:
> For example run slc loobpack to create an app called x.x outside it, it works fine, this.env.cwd returns DIR_PATH/x.x/x-x .
But run slc loobpack inside sandbox folder, it should return sandbox/x.x/x-x but it actually returns sandbox.

@jannyHou could this be caused by the code here? https://github.com/strongloop/generator-loopback/blob/8fbdb0da7ce203d387ee00096f7cfade37a0d3d9/lib/actions.js#L22-L57

jannyHou:
@bajtos I think it's either caused by `configureDestinationDir` you mentioned above or [`this line in `initWorkspace``](https://github.com/strongloop/generator-loopback/blob/8fbdb0da7ce203d387ee00096f7cfade37a0d3d9/lib/actions.js#L69)

bajtos:
@jannyHou is this patch still relevant? should we close it?

jannyHou:
@bajtos sure i am closing it now.

PR #187, opened 2016-05-20
fix/update-to-promise-support

jannyHou:
Connect to strongloop/loopback#2292

Contains relation generator fix only.

`yeoman-environment` uses the latest `inquirer`, which depends on a new module `rx` that supports promise only. So we cannot use callback in function [`checkRelationName`](https://github.com/strongloop/generator-loopback/blob/master/relation/index.js#L123)

Please review. Thanks.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -4,6 +4,7 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
+global.Promise = require('bluebird');
 var assert = require('assert');
 var chalk = require('chalk');
 var yeoman = require('yeoman-generator');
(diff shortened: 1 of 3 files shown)


lib/helpers.js
```diff
@@ -98,20 +98,18 @@ exports.validateOptionalName = function (name) {
  * Check if relation has a different name from properties
  * @param {Object} modelDefinition The model which has the relation
  * @param {String} name The user input
- * @param {Function(String|Boolean)} Callback to receive the check result.
+ * @returns {String|Boolean} Return the check result.
```
bajtos: Please add `@promise` annotation for strong-docs tooling.

jannyHou: hmm...i might be wrong, but I think the return value is not a promise, it's just a pure String/Boolean value right?

bajtos: My understanding is that returning a promise (instead of calling a callback) is the whole point of this patch/change, right?

In the code, you do `return getAsync().then(..)`, which returns a promise.

PR #188, opened 2016-05-24
remote-method: make prompts more descriptive

bajtos:
I had a discussion with another IBMer, a newbie to LoopBack, about creating models with custom remote methods. He was adding accepts/returns arguments by manually editing JSON files, because he did not realise the generator prompts were asking for arguments.

In this patch, I have reworded prompts and added more explanatory texts to guide new developers.

@richardpringle @crandmck @jannyHou please review

Here is a copy of the new generator output:

```
$ yo loopback:remote-method
? Select the model: Car
? Enter the remote method name: honk
? Is Static? Yes
? Description for method: make some noise!

Let's configure where to expose your new method in the public REST API.
You can provide multiple HTTP endpoints, enter an empty path when you are done.
? Enter the path of this endpoint: /honk
? HTTP verb: post

Let's add another endpoint, enter an empty name when done.
? Enter the path of this endpoint:

Now it's time to describe input arguments of your remote method
In LoopBack, we call these arguments "accepts" arguments.
Of course, you can define multiple input arguments.
Just enter an empty name when all arguments were defined.
? What is the name of this argument? mood
? Select argument's type: string
? Is this argument required? No
? Please describe the argument: What is driver's mood?
? Where to get the value from? (auto)

Let's add another accept argument, enter an empty name when done.
? What is the name of this argument?

Let's describe output of the new method: the arguments passed  to the callback function.
In LoopBack, we call these arguments "returns" arguments.
You can define multiple callback arguments too.
Just enter an empty name when all arguments were defined.
? What is the name of this argument? sound
? Select argument's type: buffer
? Is this argument a full response body (root)? No
? Please describe the argument: The sound in .wav format.

Let's add another return argument. Enter empty name when done.
? What is the name of this argument?

We have added strong-remoting metadata for your new method to common/models/car.json
Now it's up to you to provide implementation in common/models/car.js
Here is a sample code to get you started:

/**
 * make some noise!
 * @param {string} mood What is driver's mood?
 * @param {Function(Error, buffer)} callback
 */

Car.honk = function(mood, callback) {
  var sound;
  // TODO
  callback(null, sound);
};
```

### DIFF
--- a/remote-method/index.js
+++ b/remote-method/index.js
@@ -124,13 +124,16 @@ module.exports = yeoman.generators.Base.extend({
   },
 
   delimEndpoints: function() {
-    this.log('Let\'s add some HTTP Endpoints now.\n');
+    this.log();
+    this.log('Let\'s configure where to expose your new method ' +
+      'in the public REST API.');
+    this.log('You can provide multiple HTTP endpoints, enter an empty path ' +
+      'when you are done.');
     this.http = [];
   },
 
   askForEndpoints: function() {
     var done = this.async();
-    this.log('Enter an empty path when done.');
 
     var prompts = [
       {
@@ -171,25 +174,29 @@ module.exports = yeoman.generators.Base.extend({
           path: httpPath,
           verb: answers.httpVerb || answers.customHttpVerb
         });
-        this.log('\nLet\'s add another endpoint.');
+        this.log(
+          '\nLet\'s add another endpoint, enter an empty name when done.');
         this.askForEndpoints();
       }.bind(this));
     }.bind(this));
   },
 
   delimAccepts: function() {
-    this.log('Let\'s add some Accept Arguments now.\n');
+    this.log();
+    this.log('Describe the input ("accepts") arguments of your remote method.');
+    this.log('You can define multiple input arguments.');
+    this.log('Enter an empty name when you\'ve defined all input arguments.');
+
     this.accepts = [];
   },
 
   askForAccepts: function() {
     var done = this.async();
-    this.log('Enter an empty name when done.');
 
     var prompts = [
       {
         name: 'acceptsArg',
-        message: 'Enter name for accept argument:',
+        message: 'What is the name of this argument?',
         required: true,
         validate: validateOptionalName
       }
@@ -201,24 +208,24 @@ module.exports = yeoman.generators.Base.extend({
       var subprompts = [
         {
           name: 'acceptsType',
-          message: 'The type of accept argument:',
+          message: 'Select argument\'s type:',
           type: 'list',
           choices: typeChoices
         },
         {
           name: 'acceptsRequired',
-          message: 'Required?',
+          message: 'Is this argument required?',
           type: 'confirm',
           default: false
         },
         {
           name: 'acceptsDes',
-          message: 'Description of accept argument:',
+          message: 'Please describe the argument:',
           required: true
         },
         {
           name: 'httpSource',
-          message: 'Mapping to http source:',
+          message: 'Where to get the value from?',
           type: 'list',
           choices: [
             {
@@ -261,25 +268,29 @@ module.exports = yeoman.generators.Base.extend({
           entry.http = { source: answers.httpSource };
         }
         this.accepts.push(entry);
-        this.log('\nLet\'s add another accept argument.');
+        this.log('\nLet\'s add another accept argument, ' +
+          'enter an empty name when done.');
         this.askForAccepts();
       }.bind(this));
     }.bind(this));
   },
 
   delimReturns: function() {
-    this.log('Let\'s add some Return Arguments now.\n');
+    this.log();
+    this.log('Describe the output ("returns") arguments ' +
+     'to the remote method\'s callback function.');
+    this.log('You can define multiple output arguments.');
+    this.log('Enter an empty name when you\'ve defined all output arguments.');
     this.returns = [];
   },
 
   askForReturns: function() {
     var done = this.async();
-    this.log('Enter an empty name when done.');
 
     var prompts = [
       {
         name: 'returnsArg',
-        message: 'Enter name for return argument:',
+        message: 'What is the name of this argument?',
         required: true,
         validate: validateOptionalName
       }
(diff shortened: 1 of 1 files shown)


richardpringle:
@bajtos, I honestly don't think that a cli is the best place to describe what the `accepts` and `returns` arguments are. That's why we have documentation. I always refer to the [API docs](https://apidocs.strongloop.com/strong-remoting/#sharedmethod) when describing remoting-metadata. 

I do however like the fact that

jannyHou:
@bajtos I partly agree with @richardpringle that the prompt description might be better to keep concise. Think of the time that we don't have a remote-method generator, users still need to read doc to see the definition of each field (I mean like what is `endpoint`).

For a quick understanding of what is `accept` and `return`,
how about at the beginning when generator invoked, we add a dummy example to show how a remote-method Object looks like?
For example:
```js
/**
 * Description field of remote-method
 * @param {string} acceptArgName My first accept argument
 * @param {Function(Error, string)} callback
 */

ModelName.remoteMethodName = function(acceptArgName, callback) {
  var returnArgName;
  // TODO
  callback(null, returnArgName);
};
```
So at the beginning we tell users what  `accept`, `return`, ect are and finally we show them what they create.

crandmck:
> I honestly don't think that a cli is the best place to describe what the accepts and returns arguments are. 

I tend to agree, but I think a little more description than what we have now is appropriate.  Instead of
> Now it's time to describe input arguments of your remote method
In LoopBack, we call these arguments "accepts" arguments.
Of course, you can define multiple input arguments.
Just enter an empty name when all arguments were defined.

I suggest:
```
Describe the input ("accepts") arguments of your remote method
You can define multiple input arguments.
Enter an empty name when you've defined all the input arguments.
```

Instead of 
> Let's describe output of the new method: the arguments passed  to the callback function.
In LoopBack, we call these arguments "returns" arguments.
You can define multiple callback arguments too.
Just enter an empty name when all arguments were defined.

I suggest:

```
Describe the output ("returns") arguments to the remote method's callback function.
You can define multiple callback arguments too.
Enter an empty name when you've defined all output arguments.
```

bajtos:
Thank you all for thoughtful comments!

> I honestly don't think that a cli is the best place to describe what the accepts and returns arguments are. That's why we have documentation. I always refer to the API docs when describing remoting-metadata. 

In my experience, most people (myself included) don't read documentation. IMO, a well designed tool should guide you to be successful even without reading the manual. See [Pit of Success](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

In the particular case of `yo loopback:remote-method`, it seemed to me that the prompts asking for arguments did not make it clear what exactly they are asking about. I.e. a newcomer may not realize the prompts are asking for method inputs and then outputs.

In that light, I like @crandmck's proposal in https://github.com/strongloop/generator-loopback/pull/188#issuecomment-221339855 and will apply his suggestions.

> how about at the beginning when generator invoked, we add a dummy example to show how a remote method looks like?

I am not sure about this, it seems like printing too much text in the beginning. Perhaps we can print URL pointing to the documentation page describing remote methods instead? @crandmck @richardpringle what's your opinion?

bajtos:
Applied, PTAL again.

Here is the new output:

```
$ yo loopback:remote-method
? Select the model: Car
? Enter the remote method name: honk
? Is Static? Yes
? Description for method: make some noise!

Let's configure where to expose your new method in the public REST API.
You can provide multiple HTTP endpoints, enter an empty path when you are done.
? Enter the path of this endpoint: /honk
? HTTP verb: post

Let's add another endpoint, enter an empty name when done.
? Enter the path of this endpoint:

Describe the input ("accepts") arguments of your remote method.
You can define multiple input arguments.
Enter an empty name when you've defined all input arguments.
? What is the name of this argument? mood
? Select argument's type: string
? Is this argument required? No
? Please describe the argument: what is driver's mood?
? Where to get the value from? (auto)

Let's add another accept argument, enter an empty name when done.
? What is the name of this argument?

Describe the output ("returns") arguments to the remote method's callback function.
You can define multiple output arguments.
Enter an empty name when you've defined all output arguments.
? What is the name of this argument? sound
? Select argument's type: buffer
? Is this argument a full response body (root)? Yes
? Please describe the argument: the sound in .wav format

Let's add another return argument. Enter empty name when done.
? What is the name of this argument?

We added strong-remoting metadata for your new method to common/models/car.json.
You must implement the method in common/models/car.js. For example:
Here is a sample code to get you started:

/**
 * make some noise!
 * @param {string} mood what is driver's mood?
 * @param {Function(Error, buffer)} callback
 */

Car.honk = function(mood, callback) {
  var sound;
  // TODO
  callback(null, sound);
};
```

remote-method/index.js
```diff
@@ -166,25 +169,29 @@ module.exports = yeoman.generators.Base.extend({
...
+          '\nLet\'s add another endpoint, enter an empty name when done.');
         this.askForEndpoints();
       }.bind(this));
     }.bind(this));
   },
 
   delimAccepts: function() {
-    this.log('Let\'s add some Accept Arguments now.\n');
+    this.log();
+    this.log('Describe the input ("accepts") arguments of your remote method.');
```
richardpringle: @bajtos, I think that the brackets combined with the double quotes is a little redundant. I understand why they are both there but I think it would look at little cleaner with `the "accepts" input arguments` or possibly `the input arguments ("accepts" array)` or something along those lines. 

I'm being picky here but I think that the bracket/quote combo draws too much attention.

bajtos: Frankly, I don't really mind either way. @crandmck do you have any opinion?

richardpringle:
>In my experience, most people (myself included) don't read documentation. IMO, a well designed tool should guide you to be successful even without reading the manual.

I mean, generally you have to read the API documentation to use any framework... You can only "guess" your along so far before something fails and you have to go back to the API-docs.

jannyHou:
New messages LGTM.

crandmck:
I think saying 
`Describe the input arguments ("accepts" array) of your remote method.` 
`Describe the output arguments ("returns" array) of your remote method.`
is OK.  That's be slightly different than @richardpringle's suggestion, but I think addresses his concern.

Also, one nitpick:
`Here is a sample code to get you started:`
s/be
`Here is sample code to get you started:`

PR #190, opened 2016-05-26
Fix failing relation name test

eddiemonge:
The tests were failing because there wasn't a `getAsync` method on the mocked out object

### DIFF
--- a/test/helpers.test.js
+++ b/test/helpers.test.js
@@ -12,6 +12,7 @@ var validateRequiredName = helpers.validateRequiredName;
 var checkRelationName = helpers.checkRelationName;
 require('chai').should();
 var expect = require('chai').expect;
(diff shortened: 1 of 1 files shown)


jannyHou:
@eddiemonge Thank you and your fix of the mock looks good. Actually most of failures(40+) are caused by a upgrading of yeoman-generator dependancy. We are trying different ways to see how to fix it, like Candy's pr and another one of mine.

eddiemonge:
Split this change out to its own commit/pr as the other commit wasn't related.

This was a failure unrelated to the upgrade.

jannyHou:
@eddiemonge i run test with your fix on local and it passes.
LGTM

bajtos:
LGTM, thank you for the contribution!

PR #194, opened 2016-06-03
Fix model generator

jannyHou:
A fix to https://github.com/strongloop/loopback/issues/2292#issuecomment-223397865

Model generator is broken due to the name check of property:

```
`? Enter the model name: Contact
? Select the data-source to attach undefined to: db (memory)
? Select model's base class PersistedModel
? Expose Contact via the REST API? Yes
? Custom plural form (used to build REST URL): 
? Common model or server only? common
Let's add some Contact properties now.

Enter an empty property name when done.
? Property name: 
>> Name cannot contain special characters [object Object]email
`
```

If we use 
[`validate: validateOptionalName](https://github.com/strongloop/generator-loopback/blob/master/model/index.js#L224)
to validate name, it will provide a second argument to the function `validateOptionalName`:
use console.log to show it
```
JSON.stringify(second arg): {}
second arg: [object Object]
```
So it refuse name that match [object Object]
.

Use 
```js
validate: function(value) {
  validateOptionalName(value);
}
```
instead  to fix it by forcing the second argument as undefined.

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -221,7 +221,9 @@ module.exports = yeoman.generators.Base.extend({
       {
         name: 'propertyName',
         message: 'Property name:',
-        validate: validateOptionalName
+        validate: function(value){
+          return validateOptionalName(value);
+        }
       }
     ];
     this.prompt(prompts, function(answers) {
--- a/relation/index.js
+++ b/relation/index.js
@@ -126,7 +126,9 @@ module.exports = yeoman.generators.Base.extend({
       {
         name: 'foreignKey',
         message: 'Optionally enter a custom foreign key:',
-        validate: validateOptionalName
+         validate: function(value){
+          return validateOptionalName(value);
+        }
       },
       {
         name: 'through',
--- a/remote-method/index.js
+++ b/remote-method/index.js
@@ -198,7 +198,9 @@ module.exports = yeoman.generators.Base.extend({
         name: 'acceptsArg',
         message: 'What is the name of this argument?',
         required: true,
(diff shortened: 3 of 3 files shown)


jannyHou:
Our test case unable to cover mocking the property test in a model generator.
So I manually verify it on my local:

1. reinstall your generator-loopback
2. apply this fix into your generator-loopback code: https://github.com/strongloop/generator-loopback/pull/194/files
3. Try `slc loopback:model` and after input model info, name a property as `email`

bajtos:
This change looks rather puzzling to me. What is the underlying problem? Is inquirer providing more than one argument to `validate`? Please check all other places that are calling `validateOptionalName`, actually any of the `validate*` methods to verify they are not subject to the same issue.

Also please put more details in the commit message to explain what you are actually fixing and why.

jannyHou:
@bajtos Sure I will take a look of other places. I find the solution by chance and it's puzzling to me too.

> inquirer providing more than one argument to validate


i also suspect it.

raymondfeng:
It seems that yeoman validate is now called with two args (value and answers). See https://github.com/SBoudrias/Inquirer.js/blob/master/lib/prompts/base.js#L89

raymondfeng:
I wonder if it's better to fix https://github.com/strongloop/generator-loopback/blob/master/lib/helpers.js#L104.

jannyHou:
Closing it since https://github.com/strongloop/generator-loopback/pull/195 merged 👍

PR #197, opened 2016-06-06
Add prompt for lbVersion

jannyHou:
Connect to strongloop/loopback#2344

Add a new prompt called `loopbackVersion` and pass it into [`createFromTemplate`](https://github.com/strongloop/generator-loopback/blob/master/app/index.js#L172)

```
 {
        name: 'loopbackVersion',
        message: 'Which version of loopback do you want to install?',
        default: 2.x,
        choices: ['2.x', '3.x'],
}
```

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -153,6 +153,23 @@ module.exports = yeoman.generators.Base.extend({
     });
   },
 
+  askForLBVersion: function() {
+    var cb = this.async();
+    var prompts = [{
+      name: 'loopbackVersion',
+      message: 'Which version of LoopBack would you like to use?',
+      type: 'list',
+      default: '2.x',
+      choices: ['2.x', '3.x']
+    }];
+
+    var self = this;
+    this.prompt(prompts, function(answers) {
+      self.options.loopbackVersion = answers.loopbackVersion;
+      cb();
+    });
+  },
+
   initWorkspace: actions.initWorkspace,
 
   detectExistingProject: function() {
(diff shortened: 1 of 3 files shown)


jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@bajtos Please review. Thanks.
I will look into the failure tomorrow. My test cases pass but two swagger test cases fail.

app/index.js
```diff
@@ -153,6 +153,23 @@ module.exports = yeoman.generators.Base.extend({
     });
   },
 
+  askForLBVersion: function() {
+    var cb = this.async();
+    var prompts = [{
+      name: 'loopbackVersion',
+      message: 'Which version of loopback do you use?',
```
bajtos:
```
Which version of LoopBack would you like to use?
```

app/index.js
```diff
@@ -168,11 +185,11 @@ module.exports = yeoman.generators.Base.extend({
 
   project: function() {
     var done = this.async();
-
```
bajtos: Please keep the empty line around.

test/app.test.js
```diff
@@ -126,6 +127,22 @@ describe('loopback:app generator', function() {
     });
   });
 
+  it('scaffolds 3.x app when option.loopbackVersion is 3.x',
```
bajtos: Please add a test for `loopbackVersion: '2.x'` too, so that when we switch the default to 3.x, we will have a test for a non-default value around.

bajtos:
@jannyHou looks mostly good 👍  I left few minor comments to address, no further review is needed.

PR #198, opened 2016-06-07
Fix "undefined" model name

rundef:
A small bug I noticed when I tried to create a model ...

Before:
before

After:
after


slnode:

jannyHou:
@slnode test please

jannyHou:
@rundef Thanks for your contribution! Nice catch.
@bajtos I reproduced the error above. A minor fix. Please review.

PR #199, opened 2016-06-08
Support custom connector

jannyHou:
Connect to strongloop/loopback#2265

Now when we choose a custom connector that not included in available connectors, it returns error.
This pr is created for supporting custom connector, and print a message to remind user manually add config then.

### DIFF
--- a/datasource/index.js
+++ b/datasource/index.js
@@ -198,10 +198,13 @@ module.exports = yeoman.generators.Base.extend({
 
   installConnector: function() {
     var connector = this.availableConnectors[this.connector];
-    var pkg = connector.package;
-    if (!pkg) return;
-
-    var npmModule = pkg.name;
+    var pkg = {};
+    if (connector) {
+      pkg = connector.package;
+      if (!pkg) return;
+    }
+    
+    var npmModule = pkg.name || this.connector;
     if (pkg.version) {
       npmModule += '@' + pkg.version;
     }
@@ -244,6 +247,17 @@ module.exports = yeoman.generators.Base.extend({
       helpers.reportValidationError(err, this.log);
       return done(err);
     }.bind(this));
+  },
+
+  printAddConfigForCustomConnector: function() {
+    var connector = this.connector;
+    if (!this.availableConnectors[connector]) {
+      this.log('Please manually add config for your custom connector ' +
+        connector +
+        ' in server/datasources.json');
+    } else {
+      return;
+    }
   },
 
   saveProject: actions.saveProject
--- a/test/datasource.test.js
+++ b/test/datasource.test.js
@@ -80,6 +80,24 @@ describe('loopback:datasource generator', function() {
     });
   });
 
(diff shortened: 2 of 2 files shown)


jannyHou:
@bajtos Please review, thanks:)

jannyHou:
@slnode test please

jannyHou:
@slnode test please

test/datasource.test.js
```diff
@@ -80,6 +80,24 @@ describe('loopback:datasource generator', function() {
     });
   });
 
+  it('should support custom connector', function(done) {
+    var modelGen = givenDataSourceGenerator();
+    helpers.mockPrompt(modelGen, {
+      name: 'test-custom',
+      customConnector: 'loopback-component-storage',
```
bajtos: Please use a name that will never become a connector, e.g. `async` or `lodash`. I think the storage component is sort of connector-like.

datasource/index.js
```diff
@@ -198,7 +198,14 @@ module.exports = yeoman.generators.Base.extend({
...
-    var pkg = connector.package;
+    var pkg;
+    if (!connector) {
+      pkg = {
+        name: this.connector
+      };
+    } else {
+      pkg = connector.package;
+    }
     if (!pkg) return;
```
bajtos: While we are changing this code, can we also add support for installing packages for connectors that do not provide `package` entry? I mean when no package name is provided, we should `npm install` the connector name. This should apply both for known connectors and unknown connectors too.

bajtos:
@jannyHou reviewed, PTAL

test/datasource.test.js
jannyHou: Right. "a name that will never become a connector" makes sense.

jannyHou: `async` is already there, i use `lodash` then

bajtos:
LGTM 👏  Please rebase on top of the current master and squash the commits.

datasource/index.js
```diff
@@ -198,10 +198,13 @@ module.exports = yeoman.generators.Base.extend({
...
-    if (!pkg) return;
-
-    var npmModule = pkg.name;
+    var pkg = {};
+    if (connector) {
+      pkg = connector.package;
+      if (!pkg) return;
+    }
+    
+    var npmModule = pkg.name || connector || this.connector;
```
jannyHou: @bajtos When I check the code again, I think it should be 
`var npmModule = pkg.name || this.connector` instead of 
`var npmModule = pkg.name || connector || this.connector`

The `connector` is either undefined or an object, not a string. Thought?

bajtos: > The connector is either undefined or an object, not a string. Thought?

Ah, that's actually expected, see L200:

```js
var connector = this.availableConnectors[this.connector];
```

> I think it should be 
> ```js
var npmModule = pkg.name || this.connector;
```

Makes sense to me! 👍

jannyHou:
@slnode test please

Issue #185, opened 2016-05-18
Property default value type is incorrect


0candy:
slc loopback
then slc loopback:model
The default value prompt for properties

https://gist.github.com/cfjedimaster/74d0934d4f9f687314c531ab3387fadc
The property is of type "number", but the default value in JSON is of type "string"

Need to fix generator-loopback to convert the default value from string (produced by CLI/yeoman/inquirer prompt) to the correct type. This applies to booleans too, possibly to other types as well.

---

To clarify the scope (see https://github.com/strongloop/generator-loopback/issues/185#issuecomment-225201876 and https://github.com/strongloop/generator-loopback/issues/185#issuecomment-225205571):

 1) When the user creates a model property and provides a default value, then the swagger produced by `apic loopback:refresh` should be valid.

2) The user experience should not be made worser that it is now, this includes users calling the generator via `yo` and `slc` where we don't generate/validate swagger spec.

The current proposal in #200 throws an error when the default value cannot be parsed into property type. Now imagine the case when the user is creating a new model with 10 properties, and the 10th property aborts the generator. That's not acceptable to me.

The coercion of the default value should be implemented in such way, that as little of user input is discarded as possible.

bajtos:
@alexpit to clarify the scope:

 1) When the user creates a model property and provides a default value, then the swagger produced by `apic loopback:refresh` should be valid.

2) The user experience should not be made worser that it is now, this includes users calling the generator via `yo` and `slc` where we don't generate/validate swagger spec.

bajtos:
Elaborating a bit more on the UX part.

The current proposal in #200 throws an error when the default value cannot be parsed into property type. Now imagine the case when the user is creating a new model with 10 properties, and the 10th property aborts the generator. That's not acceptable to me.

The coercion of the default value should be implemented in such way, that as little of user input is discarded as possible.

alexpit:
Fixed in PR #200

PR #200, opened 2016-06-08
Validate the default-value type of a model property

alexpit:
connect to #185

### DIFF
--- a/property/index.js
+++ b/property/index.js
@@ -12,6 +12,7 @@ var helpers = require('../lib/helpers');
 var validateRequiredName = helpers.validateRequiredName;
 var checkPropertyName = helpers.checkPropertyName;
 var typeChoices = helpers.getTypeChoices();
+var debug = require('debug')('loopback:generator:property');
 
 module.exports = yeoman.Base.extend({
   // NOTE(bajtos)
@@ -67,6 +68,7 @@ module.exports = yeoman.Base.extend({
         name: 'name',
         message: 'Enter the property name:',
         validate: checkPropertyName,
+        default: this.propDefinition && this.propDefinition.name,
         when: function() {
           return !this.name && this.name !== 0;
         }.bind(this)
@@ -75,6 +77,9 @@ module.exports = yeoman.Base.extend({
         name: 'type',
         message: 'Property type:',
         type: 'list',
+        default: this.propDefinition && 
+          (Array.isArray(this.propDefinition.type) ? 
+            'array' : this.propDefinition.type),
         choices: typeChoices
       },
       {
@@ -90,6 +95,10 @@ module.exports = yeoman.Base.extend({
         name: 'itemType',
         message: 'The type of array items:',
         type: 'list',
+        default: this.propDefinition && 
+          this.propDefinition.type &&
+          Array.isArray(this.propDefinition.type) &&
+          this.propDefinition.type[0],
         choices: typeChoices.filter(function(t) { return t !== 'array'; }),
         when: function(answers) {
           return answers.type === 'array';
@@ -112,10 +121,17 @@ module.exports = yeoman.Base.extend({
       {
          name: 'defaultValue',
          message: 'Default value[leave blank for none]:',
-         default: null
+         default: null,
+         when: function(answers) {
+          return answers.type !== null && 
+            answers.type !== 'buffer' &&
+            answers.type !== 'any' &&
+            typeChoices.indexOf(answers.type) !== -1;
+        }
       }
     ];
     return this.prompt(prompts).then(function(answers) {
+      debug('answers: %j', answers);
       this.name = answers.name || this.name;
       if (answers.type === 'array') {
         var itemType =  answers.customItemType || answers.itemType;
@@ -123,42 +139,134 @@ module.exports = yeoman.Base.extend({
       } else {
         this.type = answers.customType || answers.type;
       }
-      this.required = answers.required;
-      this.defaultValue = answers.defaultValue;
+
+      this.propDefinition = {
+        name: this.name,
+        type: this.type,
+      };
+      if (answers.required) {
+        this.propDefinition.required = Boolean(answers.required);
+      }
+
+      if (answers.defaultValue === '') {
+          this.log('Warning: please enter the ' + this.name +
+            ' property again. The default value provided "' +
+            answers.defaultValue + 
+            '" is not valid for type: ' + this.type);
+          return this.askForParameters();
+      }
+
+      try {
+        debug('property definition input: %j', this.propDefinition);
+        coerceDefaultValue(this.propDefinition, answers.defaultValue);
+        debug('property coercion output: %j', this.propDefinition);
+      } catch (err) {
+        debug('Failed to coerce property default value: ', err);
+        this.log('Warning: please enter the ' + this.name +
+          ' property again. The default value provided "' +
+          answers.defaultValue + 
+          '" is not valid for the selected type: ' + this.type);
+        return this.askForParameters();
+      }
     }.bind(this));
   },
 
   property: function() {
     var done = this.async();
-    var def = {
-      name: this.name,
-      type: this.type
-    };
-    if (this.required) {
-      def.required = true;
-    }
-    if (this.defaultValue) {
-      if (this.type === 'boolean'){
-        if (['true', '1', 't'].indexOf(this.defaultValue) !== -1 ){
-          def.default = true;
-        } else {
-          def.default = false;
-        }
-      } else if (this.defaultValue === 'uuid' || this.defaultValue === 'guid'){
-         def.defaultFn = this.defaultValue;
-      } else if ((this.type === 'date' || this.type === 'datetime') &&
-                  this.defaultValue.toLowerCase() === 'now'){
-         def.defaultFn = 'now';
-      } else {
-         def.default = this.defaultValue;
-      }
-    }
-
-    this.modelDefinition.properties.create(def, function(err) {
+    this.modelDefinition.properties.create(this.propDefinition, function(err) {
       helpers.reportValidationError(err, this.log);
       return done(err);
     }.bind(this));
   },
-
   saveProject: actions.saveProject
 });
+
+function coerceDefaultValue(propDef, value) {
+  var itemType;
+  if (Array.isArray(propDef.type)) {
+    itemType = propDef.type[0];
+    propDef.type = 'array';
+  }
+
+  switch (propDef.type) {
+    case 'string':
+      if (value === 'uuid' || value === 'guid'){
+        propDef.defaultFn = value;
+      } else {
+        propDef.default = value;
+      }
+      break;
+    case 'number':
+      propDef.default = castToNumber(value);
+      break;
+    case 'boolean':
+      propDef.default = castToBoolean(value);
+      break;
+    case 'date':
+      if (value.toLowerCase() === 'now'){
+        propDef.defaultFn = 'now';
+      } else {
+        propDef.default = castToDate(value);
+      }
+      break;
+    case 'array':
+      propDef.type = [itemType];
+      if (itemType === 'string') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',');
+      } else if (itemType === 'number') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(castToNumber);
+      } else if (itemType === 'boolean') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(castToBoolean);
+      } else if (itemType === 'date') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(castToDate);
+      } else {
+        propDef.default = value;
+      }
+      break;
+    case 'geopoint':
+      if (value.indexOf('lat') !== -1 && value.indexOf('lng') !== -1) {
+        propDef.default = JSON.parse(value);
+      } else {
+        var geo = value.replace(/[\s,]+/g, ',').split(',');
+        propDef.default = {};
+        propDef.default.lat = castToNumber(geo[0]);
+        propDef.default.lng = castToNumber(geo[1]);
+      }
+      break;
+    case 'object':
+      propDef.default = JSON.parse(value);
+      break;
+    default:
+      propDef.default = value;
+  }
+}
+
+function castToDate(value) {
+  var dateValue;
+  var isNumber = /^[0-9]+$/.test(value);
+  if (isNumber) {
+    dateValue = new Date(castToNumber(value));
+  } else {
+    dateValue = new Date(value);
+  }
+  return dateValue;
+}
+
+function castToNumber(value) {
+  var numberValue = Number(value);
+  if (isNaN(numberValue)) {
+    throw Error('Invalid default number value: '+ value);
+  }
+  return numberValue;
+}
+
+function castToBoolean(value) {
+  if (['true', '1', 't', 'false', '0', 'f'].indexOf(value) === -1) {
+    throw Error('Invalid default boolean value "'+ value +
+      '". Expected default values: true|false, 1|0, t|f');
+  }
+  return (['true', '1', 't'].indexOf(value) !== -1) ? true : false;
+}
--- a/test/property.test.js
+++ b/test/property.test.js
@@ -35,13 +35,33 @@ describe('loopback:property generator', function() {
       });
   });
 
-  it('adds an entry to common/models/{name}.json', function(done) {
+  it('creates number type property from large number', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'age',
+      type: 'number',
+      defaultValue: '55555555555555555555.5'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('age');
+      expect(props.age).to.eql({
+        type: 'number',
+        default: 55555555555555555555.5
+      });
+      done();
+    });
+  });
+
+  it('creates model containing boolean type', function(done) {
     var propertyGenerator = givenPropertyGenerator();
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
       name: 'isPreferred',
       type: 'boolean',
-      required: 'true',
       defaultValue: 'true'
     });
 
@@ -51,26 +71,181 @@ describe('loopback:property generator', function() {
       expect(props).to.have.property('isPreferred');
       expect(props.isPreferred).to.eql({
         type: 'boolean',
-        required: true,
         default: true
       });
       done();
     });
   });
 
-  it('creates a typed array', function(done) {
+  it('creates date type property from ISO string', function(done) {
     var propertyGenerator = givenPropertyGenerator();
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
-      name: 'list',
-      type: 'array',
-      itemType: 'string'
+      name: 'year',
+      type: 'date',
+      defaultValue: '2015-11'
     });
 
     propertyGenerator.run(function() {
       var definition = common.readJsonSync('common/models/car.json');
-      var prop = definition.properties.list;
-      expect(prop.type).to.eql(['string']);
+      var props = definition.properties || {};
+      expect(props).to.have.property('year');
+      expect(props.year).to.eql({
+        type: 'date',
+        default: '2015-11-01T00:00:00.000Z'
+      });
+      done();
+    });
+  });
+
+  it('creates date type property from number', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'year',
+      type: 'date',
+      defaultValue: '1466087191000'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('year');
+      expect(props.year).to.eql({
+        type: 'date',
+        default: new Date(1466087191000).toISOString()
+      });
+      done();
+    });
+  });
+
+  it('creates string item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'options',
+      type: 'array',
+      itemType: 'string',
+      defaultValue: 'AWD,3.2L, navigation'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('options');
+      expect(props.options).to.eql({
+        type: ['string'],
+        default: ['AWD','3.2L','navigation']
+      });
+      done();
+    });
+  });
+
+  it('creates number item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'parts',
+      type: 'array',
+      itemType: 'number',
+      defaultValue: '123456, 98765'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('parts');
+      expect(props.parts).to.eql({
+        type: ['number'],
+        default: [123456, 98765]
+      });
+      done();
+    });
+  });
+
+  it('creates boolean item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'certifications',
+      type: 'array',
+      itemType: 'boolean',
+      defaultValue: 'true,1,t,false,0,f'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('certifications');
+      expect(props.certifications).to.eql({
+        type: ['boolean'],
+        default: [true, true, true, false, false, false]
+      });
+      done();
+    });
+  });
+
+  it('creates date item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'serviceDates',
+      type: 'array',
+      itemType: 'date',
+      defaultValue: '1466087191000,2016-06-16T17:13:11.000Z'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('serviceDates');
+      expect(props.serviceDates).to.eql({
+        type: ['date'],
+        default: [new Date(1466087191000).toISOString(),
+          '2016-06-16T17:13:11.000Z']
+      });
+      done();
+    });
+  });
+
+  it('creates geopoint type property from object', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'location',
+      type: 'geopoint',
+      defaultValue: '{"lat": 55.5, "lng":44.4}'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('location');
+      expect(props.location).to.eql({
+        type: 'geopoint',
+        default: {'lat': 55.5, 'lng':44.4}
+      });
+      done();
+    });
+  });
+
+  it('creates geopoint type property from numbers', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'location',
+      type: 'geopoint',
+      defaultValue: '55.5, 44.4'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('location');
+      expect(props.location).to.eql({
+        type: 'geopoint',
+        default: {'lat': 55.5, 'lng':44.4}
+      });
       done();
     });
   });


alexpit:
@richardpringle PTAL

richardpringle:
Commit message should be something like "Validate the default-value type of a model property"... or something along those lines. In general, you should be able to tell what the commit is accomplishing without looking at the actual code (although that isn't always possible).

property/index.js
```diff
@@ -166,3 +161,62 @@ module.exports = yeoman.generators.Base.extend({
 
   saveProject: actions.saveProject
 });
+
+function coerceDefault(def, value) {
```
richardpringle: Is `def` a definition? If so, I would change it to that. Not a big deal but the more we "spell things out" in the code, the easier it is to ask a community member to make a fix for us!

alexpit: nope, that's the default object

property/index.js
```diff
@@ -166,3 +161,62 @@ module.exports = yeoman.generators.Base.extend({
...
+      case 'number':
+        def.default = Number(value);
+        break;
+      case 'boolean':
+        if (['true', '1', 't'].indexOf(value) !== -1 ){
+          def.default = true;
+        } else {
+          def.default = false;
+        }
+        break;
```
richardpringle: For this case, can you just do `Boolean(value)`?  Or are you saying that default `true` has to explicitly be `true`, `1`, or `t`? 

As a JavaScript developer, I would expect any truthy value to coerce to `true`.

@bajtos might have a good comment though as he has just dealt (or is still dealing) with a bunch of LoopBack coercion issues.

property/index.js
```diff
@@ -166,3 +161,62 @@ module.exports = yeoman.generators.Base.extend({
...
+        def.default = JSON.parse(value);
+        break;
+      case 'array':
+        def.default = value.replace(/[\s,]+/g, ',').split(',');
+        break;
+      case 'date', 'datetime':
+        if (value.toLowerCase() === 'now'){
+          def.defaultFn = 'now';
+        } else {
+          def.default = new Date(value);
```
richardpringle: Is `value` always a string? I know this came up in strong-remoting, but if `value` is of JS-type number, it will give a different date than JS-type string.

I'm not sure that we handled this before but we should deal with it now. Or tell the user that anything from the command line will be parsed as a string so they should expect the correct output accordingly.

property/index.js
```diff
@@ -166,3 +161,62 @@ module.exports = yeoman.generators.Base.extend({
...
+      default:
+        if (value === 'uuid' || value === 'guid'){
+          def.defaultFn = value;
+        } else {
+          def.default = value;
+        }
+    }
+    return def;
+  } else {
+    throw Error('Unsupported model property type ' + def.type);
```
richardpringle: should have a `:`, `type:`

test/property.test.js
```diff
@@ -12,7 +12,7 @@ var SANDBOX =  path.resolve(__dirname, 'sandbox');
 var expect = require('chai').expect;
 var common = require('./common');
 
-describe('loopback:property generator', function() {
+describe.only('loopback:property generator', function() {
```
richardpringle: "only" needs to be removed

test/property.test.js
```diff
@@ -62,6 +62,121 @@ describe('loopback:property generator', function() {
...
+      expect(props.options).to.eql({
+        type: 'array',
+        required: true,
+        default: ['AWD','3.2L','navigation']
+      });
+      done();
+    });
+  });
+
+  it('should create model containing date type', function(done) {
```
richardpringle: If we are going to support both JS types of `Number` and `String` for the default value, we need to test both cases.

richardpringle:
@alexpit, left some comments for you.

richardpringle:
There are also "legitimate" (I think...) failing tests.

test/property.test.js
alexpit: Agreed

property/index.js
alexpit: Agreed

property/index.js
alexpit: The 'value' is indeed a string always (originates in the cmd line input), and I will add a check condition to ensure that.

property/index.js
alexpit: That specific code was there previously for some reason and any of the changes proposed would regress current behaviour, that's why I just left it as is.

richardpringle:
I think there should also be a warning that whenever you input a default value for property type:
  - Date: the value entered was read as an ISO formatted string
  - Buffer: the value entered was read as a string and converted to a buffer encoded as UTF8 (default for node)
  - any: warning that the value is a string

Basically, need to make sure that the user knows that all values are input as strings and then describe how each individual case is coerced. They will be able to go and make changes in JSON if they aren't happy about the coercion but at least we handled their input in a logical way.

alexpit:
@bajtos PTAL

property/index.js
bajtos:
```
$ node
> Boolean('false')
true
```

property/index.js
```diff
@@ -141,22 +142,11 @@ module.exports = yeoman.generators.Base.extend({
...
-        }
-      } else if (this.defaultValue === 'uuid' || this.defaultValue === 'guid'){
-         def.defaultFn = this.defaultValue;
-      } else if ((this.type === 'date' || this.type === 'datetime') &&
-                  this.defaultValue.toLowerCase() === 'now'){
-         def.defaultFn = 'now';
-      } else {
-         def.default = this.defaultValue;
-      }
+      def = coerceDefault(def, this.defaultValue);
```
bajtos: IIUC, `coerceDefault` takes `def` object as an input and returns a new/modified `def` object as the return value. I find it unnecessary complex, can we please simplify to:

```js
def.default = coerceDefaultValue(this.defaultValue, this.type);
```

property/index.js
```diff
@@ -141,22 +142,11 @@ module.exports = yeoman.generators.Base.extend({
...
     }
+    debug(this.modelName+ ' property: %j', def);
```
bajtos: Incorrect spacing around `+`, should be `this.modelName + '...`

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+        break;
+      case 'any':
+        console.warn('Warning: property default value was stored as string');
+        def.default = value;
+        break;
+      default:
+    }
+    return def;
+  } else {
+    throw Error('Unsupported model property type: ' + def.type);
```
bajtos: Please use reverse logic to handle error cases like this one. It makes the code easier to read and reduces nesting level.

```js
if (!isSupported || typeof value !== 'string') {
  throw new Error(...);
}

switch(...) {
  // ...
}
```

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
 
   saveProject: actions.saveProject
 });
+
+function coerceDefault(def, value) {
+  var isSupported = typeChoices.indexOf(def.type) !== -1;
```
bajtos: It is a valid use case to provide a custom type, e.g. a Model name. I think a better variable name is `isBuiltin`.

property/index.js
bajtos: Also since it's a valid case to have a custom property type, we should not throw an exception but handle it accordingly.

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+  debug('property type "%s" >> supported: %s', def.type, isSupported);
+
+  if (typeof value === 'string' && isSupported) {
+    switch ((def.type || '').toLowerCase()) {
```
bajtos: When `def.type` is an empty string, then we have an invalid property definition. This condition should be checked earlier in the flow to prevent the user to enter such value. I think we need to add a validation that `customType` is non-empty.

property/index.js
bajtos: Oh, I see, this will not work because in some cases, we are changing `default` to `defaultFn`.

In that case we should make it explicit that `coerceDefaultValue` modifies the definition:

```js
def.default = this.defaultValue
coerceDefaultValue(def);
```

Or alternatively:

```js
setDefaultValue(def, this.defaultValue);
```

property/index.js
bajtos: As far as my understanding goes, the `def` object is a property definition in LoopBack format. I would like to see a more descriptive argument name here, as `def` can stand for both `default` and `definition`.

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+      case 'string':
+        if (value === 'uuid' || value === 'guid'){
+          def.defaultFn = value;
+        } else {
+          def.default = value;
+        }
+        break;
+      case 'number':
+        def.default = Number(value);
```
bajtos: What if the value is not a valid number, e.g. string `text`? Will we gracefully handle such error input by asking the user to enter a valid default value again?

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+        break;
+      case 'boolean':
+        if (['true', '1', 't'].indexOf(value) !== -1 ){
+          def.default = true;
+        } else {
+          def.default = false;
+        }
+        break;
+      case 'object':
+        def.default = JSON.parse(value);
```
bajtos: Ditto, what if the value is not a JSON? Will we gracefully handle such error input by asking the user to enter a valid default value again?

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+        break;
+      case 'array':
+        def.default = value.replace(/[\s,]+/g, ',').split(',');
```
bajtos: This will not work very well for array of strings. I am proposing to support two formats: JSON-encoded array, e.g. `[1,2]` or `["a,b", "c"]`; and your current comma-separated list, e.g. `1,2`.

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+        break;
+      case 'date':
+        console.warn('Warning: property default value was converted' +
+          ' from string using ISO formatting');
+        if (value.toLowerCase() === 'now'){
+          def.defaultFn = 'now';
+        } else {
+          def.default = new Date(value);
```
bajtos: Note that `new Date()` supports a wide range of different formats, not only ISO formatting. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

It would be nice to support numeric timestamps for default values too. E.g. if the value matches `/^[0-9]+$/`, then we should convert it to a number before passing it to `Date` constructor.

Notice that `new Date('0') !== new Date(0)`.

property/index.js
```diff
@@ -166,3 +156,68 @@ module.exports = yeoman.generators.Base.extend({
...
+        } else {
+          var geo = value.replace(/[\s,]+/g, ',').split(',');
+          def.default = {};
+          def.default.lat = Number(geo[0]);
+          def.default.lng = Number(geo[1]);
+        }
+        break;
+      case 'buffer':
+        console.warn('Warning: property default value was converted' + 
+          'from string using UTF8 encoding');
```
bajtos: This will make it difficult for people to provide binary data. Can we use Base64 encoding instead please? AFAIK, `Buffer` constructor supports that too.

property/index.js
bajtos: IIRC, when the user specifies `array` type, we ask them to provide `itemType` too. We should coerce array items of the default value depending on `itemType`. For example, when users selects "array of numbers" and provide default value `1,2`, then we should set the default value to `[1,2]` instead of `["1","2"]`.

test/property.test.js
```diff
@@ -62,37 +58,127 @@ describe('loopback:property generator', function() {
...
-  it('creates a typed array', function(done) {
+  it('should create model containing number type', function(done) {
     var propertyGenerator = givenPropertyGenerator();
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
-      name: 'list',
+      name: 'age',
+      type: 'number',
+      required: 'true',
+      defaultValue: '555555555555555555555.5'
```
bajtos: Unless there is an intention in the number of digits in this number, then please use the simplest numeric value that works, i.e. `0`, `1` or `1.2`.

test/property.test.js
```diff
@@ -62,37 +58,127 @@ describe('loopback:property generator', function() {
     });
   });
 
-  it('creates a typed array', function(done) {
```
bajtos: Please preserve the test case for typed array and add more tests to verify "array of numbers" at least.

bajtos:
A general note: test names should not start with `should`, see https://github.com/strongloop-internal/loopback-knowledge-base/blob/master/style-guide.md#test-naming

test/property.test.js
```diff
@@ -125,6 +207,20 @@ describe('loopback:property generator', function() {
     });
   });
 
+  it('should fail to create with unsupported type', function (done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'something',
+      type: 'unknown',
```
bajtos: I don't think this is possible to achieve via CLI UI, because the `type` is picked from a list. One can specify `customType` with arbitrary value, that's a valid use case that should be supported as I commented earlier.

bajtos:
> Will we gracefully handle such error input by asking the user to enter a valid default value again?

In chat, Alex raised concern that this proposal would increase the scope of this patch too much.

Here is my primary concern:  when the user enters invalid default value, the generator blows in they face and they have to start again, which is poor UX in my opinion.

Let's be creative then and find a different solution that will take less time to implement.

For example, we can keep the default value as the original string provided by the user, and print a warning explaining that the property definition may be invalid. Something like this:

```js
try {
  // parse the value
  return parsedValue;
} catch(err) {
  debug(..., err);
  console.warn(...);
  return originalValue;
}
```

property/index.js
alexpit: agreed

property/index.js
alexpit: agreed

property/index.js
alexpit: please provide a suggestion for "a more descriptive argument name"

property/index.js
alexpit: @bajtos: can you please clarify "handle it accordingly" with a concrete example...
There isn't much to handle when the default value passed in is cannot be parsed...

property/index.js
alexpit: aggred, will use `isBuiltin`

property/index.js
alexpit: @bajtos: This is beyond the scope of this fix and should be part of a more comprehensive story that should address the user interaction error handling. Currently, there's no such error handling impl at all to prevent unacceptable input and give end-user another chance to input correct values, based on some explicit console warnings/recommendations

property/index.js
alexpit: @bajtos: please define "gracefully handle such error" with a concrete example... there are many way to do it that takes this outside the scope of the issue (see comment above)

property/index.js
alexpit: @bajtos: again, there are many ways to do it that takes this outside the scope of the issue (see comment above), therefore a concrete suggestion of "gracefully" handling is appreciated...

property/index.js
alexpit: @bajtos: all these specifics need to be exposed to the user in some form before they input... would you agree with a console warning mechanism for this purpose ?

property/index.js
alexpit: @bajtos: again, the specifics need to be exposed to the user in some form before they input... would you agree with a console warning mechanism for this purpose ?

property/index.js
alexpit: @bajtos: I defaulted to UTF8 at @richardpringle's recommendation based on the nodejs buffer encoding default... I have some doubts Base64 is far more suited for the purpose...

@richardpringle, what do you think ?

test/property.test.js
alexpit: agreed

test/property.test.js
alexpit: yes, there's an explicit intention to ensure very large numbers are also handled

test/property.test.js
alexpit: yes, "the type is picked from a list", but that doesn't mean this code path could not eventually be exercised in the future through other means beside CLI, so I do believe having the error handling in place is safe and not have any adverse impact.

bajtos:
> > Will we gracefully handle such error input by asking the user to enter a valid default value again?

> In chat, Alex raised concern that this proposal would increase the scope of this patch too much.

On the second thought, implementing a loop when prompting for a value is pretty easy, we are already doing it in other places. See [remote-method](https://github.com/strongloop/generator-loopback/blob/f308e4818139192e34734e836928cfeca60b3303/remote-method/index.js#L135-L182) for an example.

When needs to be done:
 - move the prompt for `defaultValue` ([source](https://github.com/strongloop/generator-loopback/blob/f308e4818139192e34734e836928cfeca60b3303/property/index.js#L115-L119)) into a standalone method
 - convert/validate the value inside this new method
 - when the value is not valid, re-run the prompt again via recursion

@alexpit please let me know if you have any questions or concerns about my proposal above

property/index.js
bajtos: I think `propertyDefinition` or `propDef` should work well, I personally prefer the longer form.

property/index.js
bajtos: This is the kind of issues that we routinely fix as we discover them. 

Preventing an empty custom type is a matter of adding a 3 lines-long function to `validate` property of the `customType` prompt. Actually, the code is already there, see [property/index.js#L104](https://github.com/strongloop/generator-loopback/blob/f308e4818139192e34734e836928cfeca60b3303/property/index.js#L104).

In my understanding of the current code base (which may be wrong, you are welcome to prove me wrong!), `def.type` will be always defined, therefore there is no handle falsey `def.type` value here. If there are unit-tests that pass invalid `type` or `customType` answer, then we should fix them.

test/property.test.js
bajtos: I see. In that case, please make the test name explicitly mention that you are testing a very large default value. E.g. `should create model containing number type with very large default value`

property/index.js
bajtos: Let's keep the current implementation based on the code in master. It may not be perfect, but I think it's okay for now.

property/index.js
richardpringle: @bajtos, this error is a little redundant and is what I would like to refer to as a "fail-safe". Inquirer.js takes any input as a string so you should never hit this block. Having the block there can prevent someone from incorrectly mocking the prompt in a test case.

property/index.js
richardpringle: @bajtos, @alexpit,

I suggested UTF8 because it is the default for the `Buffer` NodeJS class. I think it's fair to assume that a LoopBack Buffer would use the same defaults as NodeJS.

If the user would like to use a different encoding, they can input the default value directly into the JSON (which is what I would assume most users would do in the first place even if they wanted UTF8).

test/property.test.js
richardpringle: @bajtos, please correct me if I'm wrong but apic uses generator-loopback programatically doesn't it?

(if the answer is anything other than a simple "yes" or "no", let's take it offline)

richardpringle:
>Here is my primary concern: when the user enters invalid default value, the generator blows in they face and they have to start again, which is poor UX in my opinion.

Completely agree, re-running the prompt makes sense.

Does the CLI actually crash or does it just skip to the next property? If it just skips to the next property, I think it would be valid to land that then improve in another story. 

@bajtos I think our review process could be streamlined. We need to go over the overall design first, then the code afterwards. It's a waste of time making so many little changes only to realize we need a refactor. A top-down review process would be much more efficient.

alexpit:
@slnode please test

alexpit:
@bajtos: PTAL

test/property.test.js
bajtos: The programatic part only sets up process environment (cwd, yeoman environment, etc.). Everything else (inquirer prompts in particular) is executed the same way as if the generator was started directly via `yo loopback`.

property/index.js
bajtos: Well, I think UTF8 encoding is not very useful when it comes to binary data, for example how do you want to provide a default value that's a JPEG or PNG image? But I can live with that limitation for now and leave this discussion out of scope of this change if you prefer.

What we need to do here is to verify what actually happens when `def.default` is a Buffer instance. I mean how is it stored in model JSON file? How will loopback-datasource-juggler treat that value, will it understand it's a buffer?

I don't have bandwidth to do this check myself now, could one of you @richardpringle or @alexpit perform the check please?

property/index.js
bajtos: > @richardpringle Having the block there can prevent someone from incorrectly mocking the prompt in a test case. 

+1 :heavy_multiplication_x: :100: for throwing a helpful error when the tests is mocking prompts incorrectly.

However, I think that's not the case we have, see this screenshot where I created a property using a custom type "float":

```
$ yo loopback:property
? Select the model: Car
? Enter the property name: mpg
? Property type: (other)
? Enter the type: float
? Required? No

$ cat common/models/car.json
{
  ...
  "properties": {
    "mpg": {
      "type": "float"
    }
  },
  ...
}
```

bajtos:
@bajtos I think our review process could be streamlined. We need to go over the overall design first, then the code afterwards. It's a waste of time making so many little changes only to realize we need a refactor. A top-down review process would be much more efficient.

Agreed, thank you for constructive feedback!

property/index.js
```diff
@@ -115,54 +116,128 @@ module.exports = yeoman.generators.Base.extend({
...
+        }
+      }
+      break;
+    case 'array':
+      if (propDef.itemType === 'string') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',');
+      } else if (propDef.itemType === 'number') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(function(item) {
+            return Number(item);
```
bajtos: Please handle the case where `item` is not a valid number.

property/index.js
```diff
@@ -115,54 +116,128 @@ module.exports = yeoman.generators.Base.extend({
...
+      }
+      break;
+    case 'geopoint':
+      if (value.indexOf('lat') !== -1 && value.indexOf('lng') !== -1) {
+        propDef.default = JSON.parse(value);
+      } else {
+        var geo = value.replace(/[\s,]+/g, ',').split(',');
+        propDef.default = {};
+        propDef.default.lat = Number(geo[0]);
+        propDef.default.lng = Number(geo[1]);
```
bajtos: Please handle the case where `geo[0]` and/or `geo[1]` are not valid numbers.

property/index.js
```diff
@@ -115,54 +116,128 @@ module.exports = yeoman.generators.Base.extend({
...
+      }
+      break;
+    case 'buffer':
+      console.warn('Warning: property default value was converted' + 
+        'from string using default node.js UTF8 encoding');
+      propDef.default = new Buffer(value);
```
bajtos: In our online discussion, we agreed to remove "default" prompt for buffer types.

property/index.js
```diff
@@ -56,7 +57,7 @@ module.exports = yeoman.generators.Base.extend({
 
     if (!this.modelDefinition) {
       var msg = 'Model not found: ' + this.modelName;
-      this.log(chalk.red(msg));
+      console.warn(chalk.red(msg));
```
bajtos: I think we can revert this back, cannot we?

property/index.js
```diff
@@ -115,54 +116,128 @@ module.exports = yeoman.generators.Base.extend({
...
+        try {
+          coerceDefaultValue(this.propDefinition, answers.defaultValue);
+          debug('property definition: %j', this.propDefinition);
+        } catch (err) {
+          debug('Failed to coerce property default value: ', err);
+          this.log('Warning: please enter the ' + this.name +
+            ' property again. The default value provided "' +
+            answers.defaultValue + 
+            '" is not valid for the selected type: ' + this.type);
+          this.askForParameters();
```
bajtos: I think this should be `return this.askForParameters()`.

alexpit:
@bajtos: PTAL and let me know if anything was missed. Thanks.

property/index.js
```diff
@@ -115,54 +118,128 @@ module.exports = yeoman.generators.Base.extend({
       {
          name: 'defaultValue',
          message: 'Default value[leave blank for none]:',
-         default: null
+         default: null,
+         when: function(answers) {
+          return answers.type !== null && 
+            answers.type !== 'buffer'&&
```
bajtos: Style: add a space character before `&&`. Same for the next line.

_(we should add eslint to this repository, it would catch issues like this one)_

property/index.js
```diff
@@ -115,54 +118,128 @@ module.exports = yeoman.generators.Base.extend({
...
       }
-      this.required = answers.required;
-      this.defaultValue = answers.defaultValue;
+
+      this.propDefinition = {
+        name: this.name,
+        type: this.type
+      };
+      if (answers.required) {
+        this.propDefinition.required = true;
```
bajtos: Nitpick: Is there any particular reason why not set `required` to `answers.required` directly? If you want to coerce to a boolean, then you can use `!!answers.required` or `Boolean(answers.required)`, but I would expect inquirer to already provide a boolean value.


```js
this.propDefinition = {
  name: this.name,
  type: this.type,
  required: this.propDefinition.required,
};
```

property/index.js
```diff
@@ -115,54 +118,128 @@ module.exports = yeoman.generators.Base.extend({
...
+      }
+
+      if (answers.defaultValue) {
+        this.propDefinition.type = answers.type;
```
bajtos: IIUC, `type` was already set on L143. Why is it needed to set it again here? Am I missing something?

bajtos: Is this because `coerceDefaultValue` expects `type` and `itemType`, while workspace API expects only `type` with array+itemType encoded inside? I find it very confusing how `propDefinition` is changing there and back.

I'll ping you on Slack to discuss.

bajtos:
Please offer the user last entered "itemType" as the default value too.

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: nums
? Property type: array
? The type of array items: number
? Required? Yes
? Default value[leave blank for none]: [1,2,3]
Warning: please enter the nums property again. The default value provided "[1,2,3]" is not valid for the selected type: number
? Enter the property name: nums
? Property type: array
? The type of array items: (Use arrow keys)
❯ string
  number
  boolean
  object
  date
  buffer
  geopoint
  any
  (other)
```

property/index.js
bajtos: Actually, your implementation is missing the "back" part.

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: nums
? Property type: array
? The type of array items: number
? Required? No
? Default value[leave blank for none]: 1,2,3
```

Here is the property definition created in `common/models/avatar.json`:

```
    "nums": {
      "type": "array",
      "itemType": "number",
      "default": [
        1,
        2,
        3
      ]
    }
```

As you see, there is `type` and `itemType` instead of `type: ['number']`.

test/property.test.js
```diff
@@ -83,7 +152,103 @@ describe('loopback:property generator', function() {
...
+      defaultValue: 'AWD,3.2L, navigation'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('options');
+      expect(props.options).to.eql({
+        type: 'array',
+        itemType: 'string',
```
bajtos: As I wrote in https://github.com/strongloop/generator-loopback/pull/200/files#r67865252, this is an incorrect property definition.

The correct one:

```js
{
  type: ['string'],
  required: true,
  // ...
}
```

test/property.test.js
```diff
@@ -83,7 +152,103 @@ describe('loopback:property generator', function() {
...
+      defaultValue: '123456, 98765'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('parts');
+      expect(props.parts).to.eql({
+        type: 'array',
+        itemType: 'number',
```
bajtos: Ditto, we don't use `itemType` in LDL (loopback definition language).

test/property.test.js
```diff
@@ -93,6 +258,7 @@ describe('loopback:property generator', function() {
                       // https://github.com/yeoman/generator/issues/600
       customItemType: '', // temporary workaround for
                           // https://github.com/yeoman/generator/issues/600
+      required: true,
```
bajtos: This looks like an unrelated change, can you revert it please?

test/property.test.js
```diff
@@ -104,7 +270,7 @@ describe('loopback:property generator', function() {
     });
   });
 
-  it('creates a defaultFn: "guid" on date fields if specified', function(done) {
+  it('creates defaultFn: "guid" on date fields', function(done) {
```
bajtos: ditto, can we revert?

test/property.test.js
```diff
@@ -114,6 +280,7 @@ describe('loopback:property generator', function() {
                       // https://github.com/yeoman/generator/issues/600
       customItemType: '', // temporary workaround for
                           // https://github.com/yeoman/generator/issues/600
+      required: true,
```
bajtos: ditto, can we revert?

test/property.test.js
```diff
@@ -125,6 +292,20 @@ describe('loopback:property generator', function() {
...
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'something',
+      type: 'unknown',
+      required: 'true',
+      defaultValue: 'X'
+    });
+
+    expect(propertyGenerator.run).to.throw(Error);
```
bajtos: When expecting a function to throw, then you should check that the expected error was thrown, so that the test fails when a different error is raised. Otherwise the code in test can trigger different error (like `TypeError: undefined is not a function`) and you will never notice.

I think it may be the case here, I would expect `propertyGenerator.run` to throw because you don't pass `propertyGenerator` as the `this` value here. If it turns out I am right, then I think you should fix this line to `expect(function() { propertyGenerator.run(); }).to.throw(/.../)`.

test/property.test.js
```diff
@@ -125,6 +292,20 @@ describe('loopback:property generator', function() {
...
+    done();
```
bajtos: Since this tests completes in the same tick, there is no need to use `done` callback. Please remove this line and remove `done` from function arguments.

bajtos:
@alexpit reviewed, PTAL at the comments above.

property/index.js
alexpit: Agreed.

property/index.js
alexpit: I'll set that propDefinition directly:
```
this.propDefinition = {
  name: this.name,
  type: this.type,
  required: Boolean(answers.required),
};
```

property/index.js
alexpit: agreed.

test/property.test.js
alexpit: Not sure where does the excerpt above is aimed, but given that `mockPrompt` doesn't accept string typed array, it's probably related to the expected type of the options param.

test/property.test.js
alexpit: I won't for now, but this string/number typed array syntax stunt is just plainly counter productive and a pain in the behind for all intents and purposes... it seems like a great compact syntax, until you try to do anything useful with the value...

test/property.test.js
alexpit: agreed.

test/property.test.js
alexpit: agreed.

test/property.test.js
alexpit: will take it out completely as no longer relevant since we only fail for number casting and that is tested anyway.

alexpit:
@bajtos PTAL, thanks.

property/index.js
```diff
@@ -115,54 +119,133 @@ module.exports = yeoman.generators.Base.extend({
...
+      }
+      break;
+    case 'array':
+      propDef.type = [itemType];
+      if (itemType === 'string') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',');
+      } else if (itemType === 'number') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(function(item) {
+            return Number(castToNumber(item));
```
bajtos: Why is it necessary to wrap the result of `castToNumber` in `Number` constructor one more time? The name `castToNumber` suggests it should already return the correct type, and the implementation of `castToNumber` does that.

I think this should be enough here: 

```js
value.replace(...).split(',')
  .map(castToNumber);
```

property/index.js
```diff
@@ -115,54 +119,133 @@ module.exports = yeoman.generators.Base.extend({
...
+      }
+      break;
+    case 'geopoint':
+      if (value.indexOf('lat') !== -1 && value.indexOf('lng') !== -1) {
+        propDef.default = JSON.parse(value);
+      } else {
+        var geo = value.replace(/[\s,]+/g, ',').split(',');
+        propDef.default = {};
+        propDef.default.lat = Number(castToNumber(geo[0]));
+        propDef.default.lng = Number(castToNumber(geo[1]));
```
bajtos: ditto (both this line and the line above)

test/property.test.js
```diff
@@ -45,7 +66,6 @@ describe('loopback:property generator', function() {
       customItemType: '', // temporary workaround for
                           // https://github.com/yeoman/generator/issues/600
       type: 'boolean',
-      required: 'true',
```
bajtos: This looks like an unrelated change to me, can you revert it?

test/property.test.js
```diff
@@ -55,13 +75,54 @@ describe('loopback:property generator', function() {
       expect(props).to.have.property('isPreferred');
       expect(props.isPreferred).to.eql({
         type: 'boolean',
-        required: true,
```
bajtos: Ditto.

bajtos:
@alexpit it seems to me that number validation for array types doesn't work well:

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: nums
? Property type: array
? The type of array items: number
? Required? No
? Default value[leave blank for none]: asd
$
```

And this is the property created in `common/models/avatar.json`:

```
    "nums": {
      "type": [
        "number"
      ],
      "default": "asd"
    }
```

PTAL

property/index.js
```diff
@@ -78,6 +80,7 @@ module.exports = yeoman.generators.Base.extend({
         name: 'type',
         message: 'Property type:',
         type: 'list',
+        default: this.propDefinition && this.propDefinition.type,
```
bajtos: When the type entered in the last run was an array, then `propDefinition.type` will be array-of-something, e.g. `['string']` and thus it needs to be converted back to "array" string here.

I was not able to verify this, because when I try to enter an invalid value for array of numbers, the property is still created - see https://github.com/strongloop/generator-loopback/pull/200#issuecomment-228074338.

property/index.js
```diff
@@ -93,6 +96,7 @@ module.exports = yeoman.generators.Base.extend({
         name: 'itemType',
         message: 'The type of array items:',
         type: 'list',
+        default: this.propDefinition && this.propDefinition.itemType,
```
bajtos: There is no `propDefinition.itemType` AFAICT, see also https://github.com/strongloop/generator-loopback/pull/200/files#r68249058

bajtos:
> it seems to me that number validation for array types doesn't work well:

Ignore that comment, I was running wrong version of the generator. I'll remove both that comment and this one.

property/index.js
```diff
@@ -70,6 +71,7 @@ module.exports = yeoman.generators.Base.extend({
         name: 'name',
         message: 'Enter the property name:',
         validate: checkPropertyName,
+        default: this.propDefinition && this.propDefinition.name,
```
bajtos: Somehow this does not seem to work for me.

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: num
? Property type: array
? The type of array items: number
? Required? Yes
? Default value[leave blank for none]: asd
Warning: please enter the num property again. The default value provided "asd" is not valid for the selected type: number
? Enter the property name:
```

No default was offered for the property name. PTAL

property/index.js
bajtos: Verified it is a real problem:

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: num
? Property type: array
? The type of array items: number
? Required? Yes
? Default value[leave blank for none]: asd
Warning: please enter the num property again. The default value provided "asd" is not valid for the selected type: number
? Enter the property name: num
? Property type: (Use arrow keys)
❯ string
  number
  boolean
  object
  array
  date
  buffer
```

property/index.js
bajtos:
```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: num
? Property type: array
? The type of array items: number
? Required? Yes
? Default value[leave blank for none]: asd
Warning: please enter the num property again. The default value provided "asd" is not valid for the selected type: number
? Enter the property name: num
? Property type: array
? The type of array items: (Use arrow keys)
❯ string
  number
  boolean
  object
  date
  buffer
  geopoint
  any
  (other)
```

property/index.js
```diff
@@ -115,54 +119,133 @@ module.exports = yeoman.generators.Base.extend({
...
       }
-      this.required = answers.required;
-      this.defaultValue = answers.defaultValue;
+
+      this.propDefinition = {
+        name: this.name,
+        type: this.type,
+      };
+      if (answers.required) {
+        this.propDefinition = Boolean(answers.required);
```
bajtos: I think I have already asked about this, but why cannot we simply set `this.propertyDefinition.defaultValue = answers.required`, no conversion needed?

Also notice you are overwriting whole `this.propDefinition` object here, that's probably not what you want?

bajtos: I think I have already asked about this, but why cannot we simply set `this.propertyDefinition.required = answers.required`, no conversion needed?

Also notice you are overwriting whole `this.propDefinition` object here, that's probably not what you want?

bajtos:
Reviewed, PTAL.

property/index.js
alexpit: absolutely.

property/index.js
alexpit: yeap, an oopsie

test/property.test.js
alexpit: That change is in fact quite related (to setting required attribute only when true, but leaving it out completely, otherwise in the generated/stored model), and even reverting would imply another apparently unrelated change to other test cases, which will now require to expect the mode to unnecessary contain `required: 'false'`...

test/property.test.js
alexpit: see above.

property/index.js
alexpit: Yeah, that line is what broke some of the behaviour related to comments above... it's supposed to be:
`+      if (answers.required) {
 +        this.propDefinition.required = Boolean(answers.required);`
I initially attempted to avoid the conditional set of the attribute, until I realised it unnecessary forced the occurrence of `require: false` all over the models, when it's certainly better to no bother at all unless is required. So, I'm strongly recommending that we avoid setting required attribute unless is relevant, or gets reset from `true`

property/index.js
alexpit: see comment below (setting this.propertyDefinition.required broke the behaviour)

property/index.js
alexpit: Well, we keep beating around the same bush: the inadequacy of using that item typed array syntax that's becoming a total nuisance to handle... even despite the fact that even inquirer handles the array structure in a type, itemType tuple...

alexpit:
@bajtos PTAL

property/index.js
```diff
@@ -61,12 +62,13 @@ module.exports = yeoman.Base.extend({
 
   askForParameters: function() {
     this.name = this.options.propertyName;
-
+
```
bajtos: Trailing whitespace, please remove. (This would be usually caught by eslint, but this project is not eslint-enabled yet.)

property/index.js
```diff
@@ -112,53 +121,158 @@ module.exports = yeoman.Base.extend({
...
+      if (itemType === 'string') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',');
+      } else if (itemType === 'number') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(function(item) {
+            return castToNumber(item);
+          });
+      } else if (itemType === 'boolean') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(function(item) {
```
bajtos: Nitpick: these three lines can be simplified to `.map(castToBoolean)`.

bajtos: I mean lines 223-225, to make myself more clear.

Instead of 

```js
.map(function(item) {
  return castToBoolean(item);
});
```

one ca write

```js
.map(castToBoolean);
```

property/index.js
```diff
@@ -112,53 +121,158 @@ module.exports = yeoman.Base.extend({
...
+            return castToBoolean(item);
+          });
+      } else if (itemType === 'date') {
+        propDef.default = value.replace(/[\s,]+/g, ',').split(',')
+          .map(function(item) {
```
bajtos: ditto, `.map(castToDate)`

property/index.js
```diff
@@ -112,53 +121,158 @@ module.exports = yeoman.Base.extend({
...
+      propDef.default = JSON.parse(value);
+      break;
+    default:
+      propDef.default = value;
+  }
+}
+
+function castToDate(value) {
+  var dateValue;
+  var isNumber = /^[0-9]+$/.test(value);
```
bajtos: Note that this does not support scientific notation, e.g. `1.2e10`, but I think that's not a big deal here.

test/property.test.js
```diff
@@ -35,13 +35,37 @@ describe('loopback:property generator', function() {
...
+
+  it('creates model containing boolean type', function(done) {
     var propertyGenerator = givenPropertyGenerator();
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
       name: 'isPreferred',
+      customType: '', // temporary workaround for
+                      // https://github.com/yeoman/generator/issues/600
+      customItemType: '', // temporary workaround for
+                          // https://github.com/yeoman/generator/issues/600
```
bajtos: Looks like these four lines (64-67) got back during rebase, please remove.

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
+      defaultValue: '1466087191000'
+    });
+
+    propertyGenerator.run(function() {
+      var definition = common.readJsonSync('common/models/car.json');
+      var props = definition.properties || {};
+      expect(props).to.have.property('year');
+      expect(props.year).to.eql({
+        type: 'date',
+        default: '2016-06-16T14:26:31.000Z'
```
bajtos: Because I read the implementation, I assume that `2016-06-16T14:26:31.000Z` is same as timestamp `1466087191000`. However, readers not familiar with the implementation won't be able to tell that.

Please make the code easier to understand, for example by
 - changing this line to `default: new Date(1466087191000).toISOString()`
 - or adding a comment explaining how is the expected value related to the provided timestamp string

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
       name: 'list',
       type: 'array',
-      itemType: 'string'
+      customType: '', // temporary workaround for
+                      // https://github.com/yeoman/generator/issues/600
+      customItemType: '', // temporary workaround for
+                          // https://github.com/yeoman/generator/issues/600
+      itemType: 'object'
```
bajtos: Looks like these four lines (64-67) got back during rebase, please remove.

bajtos: Ah, I put the comment on a wrong line, sorry for that. I meant lines 132-135, `customType` and `customItemType`.

This line 136, containing `itemType` property, is ok.

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
+  });
+
+  it('creates boolean item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'certifications',
+      type: 'array',
+      itemType: 'boolean',
+      defaultValue: 't,t'
```
bajtos: Please use at least one falsey value, to ensure we don't accidentally convert all strings to `true`.

bajtos: Even better, list all supported values (`t`, `true`, `f`, `false`, etc.) to make sure all of them are covered by a test.

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
+  });
+
+  it('creates date item typed array', function(done) {
+    var propertyGenerator = givenPropertyGenerator();
+    helpers.mockPrompt(propertyGenerator, {
+      model: 'Car',
+      name: 'serviceDates',
+      type: 'array',
+      itemType: 'date',
+      defaultValue: '1466087191000,1466097191000'
```
bajtos: Similarly here, please use both timestamp and ISO format in the default values used in the test.

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
+      expect(props.location).to.eql({
+        type: 'geopoint',
+        default: {'lat': 55.5, 'lng':44.4}
+      });
       done();
     });
   });
 
-  it('creates a defaultFn: "now" on date fields if specified', function(done) {
+  it('creates a defaultFn: "now" on date fields', function(done) {
```
bajtos: Unrelated change, please revert.

test/property.test.js
```diff
@@ -51,36 +75,215 @@ describe('loopback:property generator', function() {
...
+  it('creates a defaultFn: "now" on date fields', function(done) {  
     var propertyGenerator = givenPropertyGenerator();
     helpers.mockPrompt(propertyGenerator, {
       model: 'Car',
       name: 'created',
       type: 'date',
+      customType: '', // temporary workaround for
+                      // https://github.com/yeoman/generator/issues/600
+      customItemType: '', // temporary workaround for
+                          // https://github.com/yeoman/generator/issues/600
```
bajtos: Looks like these four lines (283-286) got back during rebase, please remove.

test/property.test.js
```diff
@@ -98,6 +301,10 @@ describe('loopback:property generator', function() {
       model: 'Car',
       name: 'uniqueId',
       type: 'string',
+      customType: '', // temporary workaround for
+                      // https://github.com/yeoman/generator/issues/600
+      customItemType: '', // temporary workaround for
+                          // https://github.com/yeoman/generator/issues/600
```
bajtos: Looks like these four lines (304-307) got back during rebase, please remove.

property/index.js
```diff
@@ -75,6 +77,9 @@ module.exports = yeoman.Base.extend({
         name: 'type',
         message: 'Property type:',
         type: 'list',
+        default: this.propDefinition && 
+          (Array.isArray(this.propDefinition.type) ? 
+            'array' : this.propDefinition.type),
```
bajtos: I am afraid this doesn't seem to work correctly, when I selected array of numbers in the first run, I am offered number type as the default in the second run (instead of array type).

```
$ yo loopback:property
? Select the model: Avatar
? Enter the property name: nums
? Property type: number
? Required? Yes
? Default value[leave blank for none]: asd,asd
Warning: please enter the nums property again. The default value provided "asd,asd" is not valid for the selected type: number
? Enter the property name: nums
? Property type: (Use arrow keys)
  string
❯ number
  boolean
  object
  array
  date
  buffer
```

bajtos: Oh, ignore this comment, I choose "number" instead of "array of numbers" in the first run, my bad. The prompts work fine AFAICT 👍

alexpit:
@bajtos: PTAL

alexpit:
@slnode test please

property/index.js
```diff
@@ -12,6 +12,7 @@ var helpers = require('../lib/helpers');
 var validateRequiredName = helpers.validateRequiredName;
 var checkPropertyName = helpers.checkPropertyName;
 var typeChoices = helpers.getTypeChoices();
+var debug = require('debug')('generator:property');
```
superkhau: `loopback:generator:property`

test/property.test.js
bajtos: @alexpit looks like this is not addressed yet. I am proposing to either create two tests for different date/time formats as you did above, or alternatively keep one tests and use different format for each item.

```js
defaultValue: '1366087191000, 2015-11-01T00:00:00.000Z'
```

alexpit:
@slnode test please

PR #202, opened 2016-06-10
Add globalization

0candy:
https://github.com/strongloop/loopback/issues/2422


app/index.js
```diff
@@ -84,7 +88,7 @@ module.exports = yeoman.generators.Base.extend({
       if (err) return done(err);
       this.templates = list.map(function(t) {
         return {
-          name: util.format('%s (%s)', t.name, t.description),
+          name: g.format('%s (%s)', t.name, t.description),
```
Setogit: No need for translation here.

app/index.js
```diff
@@ -202,7 +206,7 @@ module.exports = yeoman.generators.Base.extend({
   },
 
   generateYoRc: function() {
-    this.log('Generating .yo-rc.json');
+    g.log('Generating .yo-rc.json');
```
Setogit: In this case, two things to consider:
1. What is 'this`?  'this` in JavaScript is a reserved word and has a special meaning.  `this.log` may be `console.log` may be client specific function.  `g.log` is a `console.log` equivalent.
2. `.yo-rc.json` should not be translated.  The translator will probably figure out, but we have a notation to specify `don't translate."  We should use it here.

Probably the best globalization is:

`this.log(g.f('Generating {{.yo-rc.json}}'));`

app/index.js
```diff
@@ -24,34 +28,34 @@ module.exports = yeoman.generators.Base.extend({
...
     this.option('explorer', {
-      desc: 'Add Loopback Explorer to the project (true by default)',
+      desc: g.f('Add Loopback Explorer to the project (true by default)'),
       type: Boolean
     });
   },
 
   greet: function() {
-    this.log(yosay('Let\'s create a LoopBack application!'));
+    g.log(yosay(g.f('Let\'s create a LoopBack application!')));
```
Setogit: This things:
1. `this.log` may not be `console.log`
2. Even if `this.log` is `console.log`,   `g.log(yosay( ...` is not an error and works fine, but won't have nothing to extract, thus, no need to globalize.  Just `this.log(yosay(g.f('Let\'s create a LoopBack application!')));` is the best globalization here.

app/index.js
```diff
@@ -217,27 +221,27 @@ module.exports = yeoman.generators.Base.extend({
       this.log();
     }
 
-    this.log('Next steps:');
+    g.log('Next steps:');
```
Setogit: Same here.  `this.log` should not be globalized.  `this.log(g.f('Next steps:'));`

boot-script/index.js
```diff
@@ -4,8 +4,12 @@
 // License text available at https://opensource.org/licenses/MIT
 
 'use strict';
-var helpers = require('../lib/helpers');
+
 var path = require('path');
+var SG = require('strong-globalize');
```
Setogit: Minor comment: no need to have `require('path')` before `require('strong-globalize`)`

boot-script/index.js
```diff
@@ -33,7 +37,7 @@ module.exports = yeoman.generators.Base.extend({
 
     var question = {
       name: 'name',
-      message: 'Enter the script name (without `.js`):',
+      message: g.f('Enter the script name (without `.js`):'),
```
Setogit: Can use {{.js}}

datasource/index.js
```diff
@@ -48,8 +52,8 @@ module.exports = yeoman.generators.Base.extend({
 
       this.listOfAvailableConnectors = list.map(function(c) {
         var support = c.supportedByStrongLoop ?
-          ' (supported by StrongLoop)' :
-          ' (provided by community)';
+          g.f(' (supported by StrongLoop)') :
+          g.f(' (provided by community)');
         return {
           name: c.description + support,
```
Setogit: A minor globalization issue here is the concatenation `c.description + support`  Text concatenation with `+` operator of multiple strings may compose a right English, but not usually the case in other languages.

This particular case is not a problem.  String concatenation in code is one of the typical globalization bugs.

datasource/index.js
```diff
@@ -99,15 +103,16 @@ module.exports = yeoman.generators.Base.extend({
     var prompts = [
       {
         name: 'connector',
-        message: 'Select the connector for ' + displayName + ':',
+        message: g.f('Select the connector for %s:', displayName),
```
Setogit: Yup, this is the case of `string concatenation` globalization bug and  the right solution.

datasource/index.js
```diff
@@ -173,7 +178,7 @@ module.exports = yeoman.generators.Base.extend({
     if (!prompts.length && !warnings.length)
       return;
 
-    this.log('Connector-specific configuration:');
+    g.log('Connector-specific configuration:');
```
Setogit: `this.log`

datasource/index.js
```diff
@@ -216,7 +221,7 @@ module.exports = yeoman.generators.Base.extend({
     var prompts = [
       {
         name: 'installConnector',
-        message: 'Install ' + npmModule,
+        message: g.f('Install %s', npmModule),
```
Setogit: :+1:

remote-method/index.js
```diff
@@ -328,7 +344,7 @@ module.exports = yeoman.generators.Base.extend({
           root: answers.returnsRoot,
           description: answers.returnsDes
         });
-        this.log(
+        g.log(
```
Setogit: this.log

swagger/index.js
```diff
@@ -253,28 +257,28 @@ module.exports = yeoman.generators.Base.extend({
...
+          g.log(chalk.green(g.f('Model definition created/updated for %s.',
+            modelDef.name)));
           cb();
         }
       }
 
       if (self.selectedModels[modelDef.name] === SELECTED_FOR_UPDATE) {
-        self.log(chalk.green('Updating model definition for ' +
-          modelDef.name + '...'));
+        g.log(chalk.green(g.f('Updating model definition for %s...',
```
Setogit: self.log

Setogit:
Line comments added inline.  Two top level comments below.

1. No need to check in intl/MSG.json -- that's for debugging only.  Don't check in all the intl/*/*.json except the intl/en/*.json.  They are machine translation which can and should be generated when needed.

2. In case no place holders such as `%s`, both `g.f` and `g.t` work. but `g.t` is a bit cheaper than `g.f`

0candy:
> 2. Where is SetRootDir call?

In app/index.js.

0candy:
Failing for node 0.10 windows because of this https://github.com/strongloop/strong-globalize/issues/46

app/index.js
bajtos: IMO, we need to ensure that at least `t.description` is localized. This probably means enhancing loopback-workspace with globalization support.

.gitignore
```diff
@@ -28,3 +28,15 @@ temp/
 test/sandbox
 test/.pkgcache/
 .pkgcache/
+intl/MSG.json
```
bajtos: Instead of listing all different languages and files, can we afford to git-ignore everything in `intl` folder? 

BTW, keep in mind that npm excludes git-ignored files from the package by default, so we need extra config to bring `intl` files back when publishing to npmjs.org.

0candy: We only need the `intl/en` directory.
I will change it to
```
!intl/
intl/*
!intl/en/
```

0candy:
@jannyHou or @superkhau PTAL. Thanks.

app/index.js
```diff
@@ -24,34 +28,34 @@ module.exports = yeoman.Base.extend({
...
 
     this.option('skip-next-steps', {
-      desc: 'Do not print "next steps" info',
+      desc: g.f('Do not print "next steps" info'),
       type: Boolean,
     });
 
     this.option('explorer', {
-      desc: 'Add Loopback Explorer to the project (true by default)',
+      desc: g.f('Add Loopback Explorer to the project (true by default)'),
```
jannyHou: I personally tend to wrap {{true}} here..

app/index.js
```diff
@@ -24,34 +28,34 @@ module.exports = yeoman.Base.extend({
...
       type: Boolean,
     });
   },
 
   greet: function() {
-    this.log(yosay('Let\'s create a LoopBack application!'));
+    this.log(yosay(g.f('Let\'s create a LoopBack application!')));
```
jannyHou: And wrap {{Loopback}} here.

app/index.js
```diff
@@ -226,27 +230,27 @@ module.exports = yeoman.Base.extend({
...
-      this.log('  Change directory to your app');
+      this.log(g.f('  Change directory to your app'));
       this.log(chalk.green('    $ cd ' + this.dir));
       this.log();
     }
     if (cmd === 'apic') {
-      this.log('  Run API Designer to create, test, and publish your' +
-        ' application');
+      this.log(g.f('  Run API Designer to create, test, and publish your' +
+        ' application'));
```
jannyHou: And maybe wrap {{API Designer}} ?

datasource/index.js
```diff
@@ -93,15 +97,16 @@ module.exports = yeoman.Base.extend({
...
         type: 'list',
         default: 'memory',
         choices: connectorChoices,
       },
       {
         name: 'customConnector',
         message:
-          'Enter the connector name without the loopback-connector- prefix:',
+          g.f('Enter the connector name without the ' +
+            'loopback-connector- prefix:'),
```
jannyHou: I think {{loopback-connector-}} should be wrapped.

lib/helpers.js
```diff
@@ -31,21 +35,21 @@ exports.reportValidationError = function(err, logFn) {
...
 exports.validateAppName = function(name) {
   if (name.charAt(0) === '.') {
-    return 'Application name cannot start with .: ' + name;
+    return g.f('Application name cannot start with .: %s', name);
   }
   if (name.match(/[\/@\s\+%:]/)) {
-    return 'Application name cannot contain special characters (/@+%: ): ' +
-      name;
+    return g.f('Application name cannot contain special characters ' +
+      '(/@+%: ): %s', name);
```
jannyHou: maybe wrap {{(/@+%: ): %s}}

lib/helpers.js
```diff
@@ -31,21 +35,21 @@ exports.reportValidationError = function(err, logFn) {
...
   }
   if (name.toLowerCase() === 'node_modules') {
-    return 'Application name cannot be node_modules';
+    return g.f('Application name cannot be node_modules');
```
jannyHou: {{node_modules}}

lib/helpers.js
```diff
@@ -31,21 +35,21 @@ exports.reportValidationError = function(err, logFn) {
...
   }
   if (name.toLowerCase() === 'favicon.ico') {
-    return 'Application name cannot be favicon.ico';
+    return g.f('Application name cannot be favicon.ico');
```
jannyHou: {{favicon.ico}}

lib/helpers.js
```diff
@@ -31,21 +35,21 @@ exports.reportValidationError = function(err, logFn) {
...
   }
   if (name !== encodeURIComponent(name)) {
-    return 'Application name cannot contain special characters escaped by ' +
-      'encodeURIComponent: ' + name;
+    return g.f('Application name cannot contain special characters escaped by' +
+    ' %s encodeURIComponent: ', name);
```
jannyHou: I think `%s` should after encodeURIComponent?
should be

g.f('Application name cannot contain special characters escaped by' +
  ' encodeURIComponent:  %s', name)

lib/helpers.js
```diff
@@ -71,13 +75,13 @@ exports.checkPropertyName = function(name) {
  */
 exports.validateRemoteMethodName = function(name) {
   if (name.match(/\./) && !name.match(/^prototype\.([^.]*)$/)) {
-    return 'Name cannot contain "." characters except' +
-      ' the dot in "prototype." prefix';
+    return g.f('Name cannot contain "." characters except' +
+      ' the dot in "prototype." prefix');
```
jannyHou: {{"prototype."}}

middleware/index.js
```diff
@@ -173,7 +178,7 @@ module.exports = yeoman.Base.extend({
     var prompts = [
       {
         name: 'params',
-        message: 'Configuration parameters in JSON format:',
+        message: g.f('Configuration parameters in JSON format:'),
```
jannyHou: {{JSON}} if necessary...

property/index.js
```diff
@@ -106,21 +110,21 @@ module.exports = yeoman.Base.extend({
...
         name: 'required',
-        message: 'Required?',
+        message: g.f('Required?'),
         type: 'confirm',
         default: false,
       },
       {
         name: 'defaultValue',
-        message: 'Default value[leave blank for none]:',
+        message: g.f('Default value[leave blank for none]:'),
```
jannyHou: {{none}} if necessary...

remote-method/index.js
```diff
@@ -376,15 +395,15 @@ function buildIntroduction(def) {
   jsonFilePath = chalk.yellow(jsonFilePath);
 
   var tip = [
-    util.format(
+    g.f(
       'We added strong-remoting metadata for your new method to %s.',
```
jannyHou: {{strong-remoting}} if necessary..

swagger/index.js
```diff
@@ -347,7 +351,8 @@ module.exports = yeoman.Base.extend({
     generateApis(self, function(err) {
       if (!err) {
         self.log(
-          chalk.green('Models are successfully generated from swagger spec.'));
+          chalk.green(g.f('Models are successfully generated from ' +
+            'swagger spec.')));
```
jannyHou: {{swagger}} if necessary...

jannyHou:
@0candy I don't have confidence whether some terminologies would be wrapped or not, so I comment under some words that IMO will cause users confused if they are accidentally translated to, for example, mandarin.
If they are safe without wrap, please just ignore my picky comment lol.

0candy:
@jannyHou Thanks for reviewing. Fixed.

jannyHou:
:+1:  LGTM

PR #205, opened 2016-06-21
app: describe stability of 2.x/3.x versions

bajtos:
Improve the prompt for LoopBack version to tell the user what level of stability to expect from each version range.

I feel the current prompt does not give enough information to the user to make a good decision:

```
? Which version of LoopBack would you like to use? (Use arrow keys)
❯ 2.x
  3.x
```

Here is the new prompt:

```
? Which version of LoopBack would you like to use? (Use arrow keys)
❯ 2.x (stable)
  3.x (pre-release, alpha quality)
```

@jannyHou please review the implementation details
@ritch @crandmck @richardpringle please take a look at the additional text we are printing now and let me know how can we improve it.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -148,13 +148,29 @@ module.exports = yeoman.Base.extend({
     }.bind(this));
   },
 
+  fetchLoopBackVersions: function() {
+    var done = this.async();
+    var self = this;
+    Workspace.getAvailableLBVersions(function(err, versionsMap) {
+      if (err) return done(err);
+      var versionNames = Object.keys(versionsMap);
+      self.availableLBVersions = versionNames.map(function(version) {
+        return {
+          name: version + ' (' + versionsMap[version].description + ')',
+          value: version,
+        };
+      });
+      done();
+    });
+  },
+
   askForLBVersion: function() {
     var prompts = [{
       name: 'loopbackVersion',
       message: 'Which version of LoopBack would you like to use?',
       type: 'list',
       default: '2.x',
-      choices: ['2.x', '3.x'],
+      choices: this.availableLBVersions,
     }];
 
     var self = this;


jannyHou:
LGTM :)

bajtos:
On the second thought, as I wrote in https://github.com/strongloop/generator-loopback/pull/207#issuecomment-227672375, the description of LoopBack versions should be maintained in loopback-workspace, not in the generator. I'll rework this patch accordingly.

Having said that, I think it's find to review the description strings here until I open the pull request in loopback-workspace.

jannyHou:
@slnode test please

bajtos:
I have reworked the patch to use the new workspace API added in https://github.com/strongloop/loopback-workspace/pull/290.

@jannyHou PTAL.

bajtos:
@jannyHou I am proposing to land this patch first, as it's fixing a real usability issue, and only then land your https://github.com/strongloop/generator-loopback/pull/208 implementing version-based filtering

jannyHou:
LGTM 
Tried on local:

```
? Which version of LoopBack would you like to use? (Use arrow keys)
❯ 2.x (stable) 
  3.x (pre-release, alpha quality) 
```

richardpringle:
@bajtos @jannyHou 

I don't like "alpha quality". Theoretically, you could be so good that there are no changes from alpha to full release. "alpha" doesn't describe the quality per se. I would strongly recommend that we remove that word.

crandmck:
Perhaps instead of "alpha quality", we could say something like "not fully tested", "not tested for production" or "not certified for production" or something along those lines.  Or just leave it at "pre-release".  I think most people understand what that means.

bajtos:
To be honest, I don't really mind the exact wording, as long as it's clear that `3.0` is not ready for production yet.

@richardpringle @crandmck could you please propose what exactly should the prompt show?

 - `3.x (pre-release)`
 - `3.x (pre-release, don't use in production)`
 - something else?

bajtos:
Let's move this discussion to https://github.com/strongloop/loopback-workspace/pull/307.

I am going to land this patch, as it improves the current situation and the text change has to be done elsewhere.

PR #207, opened 2016-06-22
Apply filter on templates

jannyHou:
Connect to strongloop-internal/scrum-loopback#926

Move the version 2/3.x selection before the template prompt and filter based on it.
After controller adding to `hello-world` template, it will not support 2.x, so I delete the choice of `hello-world` when 2.x is applied.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -188,3 +188,13 @@ actions.addNullDataSourceItem = function() {
     value: null
   });
 };
+
+actions.templateFilter = function() {
+  if (this.options.loopbackVersion === '2.x') {
+    for (var template in this.templates) {
+      if (this.templates[template].name.includes('hello-world')) {
+        delete this.templates[template];
+      }
+    }
+  }
+};
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -192,7 +192,7 @@ actions.addNullDataSourceItem = function() {
 actions.templateFilter = function() {
   if (this.options.loopbackVersion === '2.x') {
     for (var template in this.templates) {
-      if (this.templates[template].name.includes('hello-world')) {
+      if (this.templates[template].name.indexOf('hello-world') > -1) {
         delete this.templates[template];
       }
     }
--- a/app/index.js
+++ b/app/index.js
@@ -135,6 +135,25 @@ module.exports = yeoman.generators.Base.extend({
 
   configureDestinationDir: actions.configureDestinationDir,
 
+  askForLBVersion: function() {
+    var cb = this.async();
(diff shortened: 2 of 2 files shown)


jannyHou:
@bajtos I didn't include your description change in pr https://github.com/strongloop/generator-loopback/pull/205, prefer to rebase on it after you merge it.

And I think I'd better merge this pr first (if possible) then implement the validation test story (another pr already created):
https://github.com/strongloop-internal/scrum-loopback/issues/902 , so the version prompt comes before template prompt.

lib/actions.js
```diff
@@ -188,3 +188,13 @@ actions.addNullDataSourceItem = function() {
     value: null
   });
 };
+
+actions.templateFilter = function() {
+  if (this.options.loopbackVersion === '2.x') {
```
bajtos: The action name templateFilter does not make it clear where it's taking the `loopbackVersion` option, it creates coupling that's difficult to spot. I'd prefer to keep this function inside the generator that's using it. I assume you had a reason for moving it to lib/actions, what was it?

lib/actions.js
```diff
@@ -188,3 +188,13 @@ actions.addNullDataSourceItem = function() {
...
+    for (var template in this.templates) {
+      if (this.templates[template].name.indexOf('hello-world') > -1) {
```
bajtos: Why are we filtering `hello-world` out for 2.x version range?

bajtos: > After controller adding to hello-world template, it will not support 2.x, so I delete the choice of hello-world when 2.x is applied.

This does not make sense to me. The hello-world template implements a controller-like model that's fully compatible with LoopBack 2.x.

bajtos:
1) As I am thinking this through, I realised the decision of which templates supports which LoopBack versions should be made in loopback-workspace. We already return extra metadata for each template, we should extend this metadata with a list of LoopBack versions each template supports. See https://github.com/strongloop/loopback-workspace/blob/c1e36891cbb71b6898ebff8f188ecb873626b10c/common/models/workspace.js#L90-L93 for starter.

2) I also think that the list of LoopBack versions available (2.x, 3.x) should not be maintained here in generator, but there should be an API in workspace for that, because the workspace knows which LoopBack versions it can scaffold.

I am proposing that you, @jannyHou, start with 1), and then we can decide who has bandwidth to address 2).

Thoughts?

jannyHou:
I am closing this pr and let's move discussion to the new prs:
loopback-workspace: https://github.com/strongloop/loopback-workspace/pull/290
generator-loopback:

PR #212, opened 2016-07-15
Replace-callback-with-promise

jannyHou:
Connect to strongloop/loopback#2505

Relation Name Check has a compatible problem with the latest yeoman generator.
Error details see:
```
jisoos-mbp:ji jisoolee$ apic loopback:relation
? Select the model to create the relationship from: Message
? Relation type: has many
? Choose a model to create a relationship with: Message
? Enter the property name for the relation: (messages) xyzErrorError: undefined
```

I believe it's caused by missing strongloop/generator-loopback@ba69e64
I apply that commit and the related test case here to fix it.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -124,20 +124,18 @@ function validateValue(name, unallowedCharacters) {
  * Check if relation has a different name from properties
  * @param {Object} modelDefinition The model which has the relation
  * @param {String} name The user input
- * @param {Function(String|Boolean)} Callback to receive the check result.
+ * @returns {String|Boolean} Return the check result.
  */
-exports.checkRelationName = function (modelDefinition, name, callback) {
-  modelDefinition.properties(function(err, list) {
-    if (err) callback(err);
-    var conflict = list.some(function(property) {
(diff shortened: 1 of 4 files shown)


jannyHou:
@slnode test please

jannyHou:
@bajtos PTAL. And one windows machine still fails due to [bug in strong-cache-install](https://github.com/strongloop/strong-cached-install/issues/9).
I am so sorry that I planned to debug it but I forgot to bring my windows laptop home :( So I will continue looking into it on Monday.

jannyHou:
And again, it's not caught by our test cases. 
We have test for relation generator and also the mock test for CheckRelationName but neither of the tests catches this error.
I will mark it in homework2 to refactor the test code in the future.

bajtos:
LGTM, :shipit: 

> And again, it's not caught by our test cases. 
We have test for relation generator and also the mock test for CheckRelationName but neither of them catches this error.
I will mark it in homework2 to refactor the test code in the future.

Sounds good, please create a GH issue to keep track of that work.

Issue #213, opened 2016-07-16
Can't set the default value of property to blank/none any more


coderluy:
apiconnect-cli-loopback test catches the issue. Looks like a regression introduced by https://github.com/strongloop/generator-loopback/pull/200

```
apic loopback:property
$ apic loopback:property
? Select the model: Note
? Enter the property name: user
? Property type: string
? Required? No
? Default value[leave blank for none]: 
Warning: please enter the user property again. The default value provided "" is not valid for type: string
? Enter the property name:
```

coderluy:
```$ apic --ext-version
apiconnect: v2.1.0 (API Connect: vdev)
 apiconnect-cli-apis 2.1.14 (dda3613ccaa5f612e1336c372bf120a8adf0f87d)
 apiconnect-cli-apps 2.1.16 (83c14e9a2e470888b5b6c7bcde974b591bd8e7ad)
 apiconnect-cli-auth 2.1.27 (71a0faa7661556658c2ffec2ef8e75ea5d9c3a64)
 apiconnect-cli-catalogs 2.1.5 (9e5737692eddfafee73686ec3d94c0610220cc33)
 apiconnect-cli-config 2.1.5 (e01f4d1578259eb357c804027dc10cc4d20a5ad0)
 apiconnect-cli-create 2.1.14 (74cf204378f885f8866065e56a5a0acfe462c25b)
 apiconnect-cli-drafts 2.1.13 (d851d11ea809498bc0919b8b78cce8818e6f4696)
 apiconnect-cli-edit 2.1.10 (c11b72d44a9e59ba96f98c92905d028d7748c42c)
 apiconnect-cli-loopback 2.1.17 (e88f0035f3182c0b3ee23ca1c682aae2db131803)
 apiconnect-cli-orgs 2.1.4 (43e1095fa047f50ceda63ad30ae44db660b5920c)
 apiconnect-cli-pm 2.1.14 (20d8fb2bdc5ee584ded04bb591c5cd16e65aed67)
  microgateway 1.2.17 (15332c13fe7554cdb28bf5f243d8a833af7c7ce0)
   flow-engine 1.1.0 (5b0fd2d88618b2dc772f711bf54d828f4af60b69)
 apiconnect-cli-policies 1.1.10 (b9ba41d63b20c72549b3112691f5a70bc0503180)
 apiconnect-cli-products 2.1.16 (25d8ccafa5a4cf6a5914a41ed5f3f3053ca077b1)
[...]

PR #216, opened 2016-08-02
Always allow properties with no default value

bajtos:
Revert the behaviour change introduced by 950c8b, as it was confusing users.

I am intentionally not including a unit test, as I am not able to create a failing one. It seems that the mock prompt is converting empty strings to nulls under the hood, which bypasses the original check `if (required && defaultValue === '')`.

Connect to #213 

@0candy or @jannyHou: please review

### DIFF
--- a/property/index.js
+++ b/property/index.js
@@ -152,11 +152,9 @@ module.exports = yeoman.Base.extend({
         this.propDefinition.required = Boolean(answers.required);
       }
 
-      if (this.propDefinition.required && answers.defaultValue === '') {
-        this.log(g.f('Warning: please enter the %s property again. ' +
-            'The default value provided "%s" is not valid for type: %s',
-            this.name, answers.defaultValue, this.type));
-        return this.askForParameters();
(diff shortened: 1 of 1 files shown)


jannyHou:
@bajtos LGTM 👍

0candy:
@bajtos This won't always allow the user to set no default value. For example, if I try to set the property value to date/boolean/array of string and others, I will get the warning `Warning: please enter the prop1 property again. The default value provided "null" is not valid for the selected type:`

Setting default value to null will cause `coerceDefaultValue` to throw a warning and ask the user to enter a default value again.

https://github.com/strongloop/generator-loopback/blob/2a6210ef769b3a37d73b9c7b4eed8677a9fc4a77/property/index.js#L155-L169

bajtos:
@0candy good catch! I fixed the issue in b8a5104. PTAL again.

0candy:
LGTM. Needs rebasing and squash commits.

PR #217, opened 2016-08-03
Missed globalizing some warnings

0candy:
* Missed globalizing a few warnings.
* Update to latest version of strong-globalize
* Fixed rootdir to use autonomousMsgLoading `all`

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -7,7 +7,7 @@
 
 var path = require('path');
 var SG = require('strong-globalize');
-SG.SetRootDir(path.join(__dirname, '..'));
+SG.SetRootDir(path.join(__dirname, '..'), {autonomousMsgLoading: 'all'});
 var g = SG();
 
 var yeoman = require('yeoman-generator');
@@ -44,7 +44,7 @@ module.exports = yeoman.Base.extend({
     });
 
     this.option('explorer', {
-      desc: g.f('Add {{Loopback Explorer }}to the project ({{true}} ' +
+      desc: g.f('Add {{Loopback Explorer}} to the project ({{true}} ' +
         'by default)'),
       type: Boolean,
     });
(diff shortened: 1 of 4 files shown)


0candy:
@jannyHou PTAL. Thanks

app/index.js
```diff
@@ -7,7 +7,7 @@
 
 var path = require('path');
 var SG = require('strong-globalize');
-SG.SetRootDir(path.join(__dirname, '..'));
+SG.SetRootDir(path.join(__dirname, '..'), {autonomousMsgLoading: 'all'});
```
jannyHou: I might miss something here, when do we need to set `autonomousMsgLoading` as `all`?

jannyHou:
@0candy Only one question about the config var `autonomousMsgLoading`, others LGTM :+1:

app/index.js
0candy: We need to add it to the entry point modules.

jannyHou: Do we need to update all repos to include it in entry point modules(like index.js in root folder), or it's only required in generator-lb?

0candy: We don't need to update all repos. Just the top modules. I will probably add it to `loopback` as well.

Amir-61: > Just the top modules. I will probably add it to loopback as well.
I'm just curious; is this needed for `loopback-datasource-juggler`, which is one of the main modules.

0candy: @Amir-61  Is it an entry point to a loopback app? Or will loopback and generator-loopback be accessed first?

Amir-61: >Is it an entry point to a loopback app?

No I don't think if `loopback-datasource-juggler` is the entry point.

I was just thinking of `loopback-datasource-juggler` based on this comment:
>We don't need to update all repos. Just the top modules. I will probably add it to loopback as well.

So if it is just for entry point modules, then I don't think if we needed it for `loopback-datasource-juggler` and we are good.

Thanks!

Issue #224, opened 2016-09-09
Using the generator to create a non-required date sets an invalid default value of `null`


samt:
Example:

```
? Enter the model name: docnull
? Select the data-source to attach docnull to: docnull (memory)
? Select model's base class PersistedModel
? Expose docnull via the REST API? Yes
? Custom plural form (used to build REST URL): 
? Common model or server only? common
Let's add some docnull properties now.

Enter an empty property name when done.
? Property name: date
   invoke   loopback:property
? Property type: date
? Required? No
? Default value[leave blank for none]: 

Let's add another docnull property.
Enter an empty property name when done.
? Property name: 
Done running LoopBack generator
```

Results in the following (excerpt) in the yaml:
```yaml
# ...
definitions:
  docnull:
    properties:
      date:
        default: null
        type: string
        format: date
      id:
        type: number
        format: double
    additionalProperties: false
  x-any:
    properties: {}
# ...
```
As null is an invalid value for a `date` type, it will fail validation is the field is omitted.

tomvanoppens:
in addition , any properties shouldn't be there when It's empty

ritch:
This looks like a bug in the swagger generation. We shouldn't be generating a default value for swagger of `null` in this case.

Actually I don't think we should ever generate the `default` field in swagger since it is the default input for the swagger client and not the value of the property. We would need to add a `swaggerDefault` or similar setting in the remoting meta.

davidcheung:
for me when i leave it empty, it does not populate the `key` for `default` in my **model.json**,
```js
$ slc loopback:property
? Select the model: farm
? Enter the property name: updated
? Property type: date
? Required? No
? Default value[leave blank for none]: 

//model.json
"updated": {
   "type": "date"
},
```
but when i put in something invalid such as an **empty space**, or something invalid, it converts it to null
```js
slc loopback:property
? Select the model: farm
? Enter the property name: updated
? Property type: date
? Required? No
? Default value[leave blank for none]: foo
  loopback:generator:property answers: {"name":"updated","type":"date","required":false,"defaultValue":"foo"} +0ms
  loopback:generator:property property definition input: {"name":"updated","type":"date"} +2ms
  loopback:generator:property property coercion output: {"name":"updated","type":"date","default":null} +3ms

//model.json
"updated": {
  "type": "date",
  "default": null
},
```

Looks like this line is whats returning 'Invalid Date', then falling back to `default:null`
[...]

PR #225, opened 2016-09-14
property: Avoid casting invalid date into `null`

davidcheung:
fixes https://github.com/strongloop/generator-loopback/issues/224
avoids scenario where generator ends up with `default: null` in date objects as they are not valid dates, it will now throw error which prompts user to enter the property again


davidcheung:
@bajtos PTAL

Issue #227, opened 2016-09-16
Generator usage files are not translated.


0candy:
Need to move usage files to `intl/en` folder, update code to read help file from that folder and get the translation strings for them.
Name the help file to `.txt`
Put `{{}}` around terms or commands that don't need to be globalized.

jannyHou:
I am afraid the only way to get translation work is **hard-coding all help texts in .txt files and remove usage file in each generator folder.**

Reason:

- Now we rely on `yeoman.base.prototype.help()` to generate the helpString, it's not only returning contents in `Usage` file but also returning other info like usage, options, arguments, etc..., which are generated dynamically. While strong-globalize needs static strings/texts.

- `yeoman.base.prototype.help()` doesn't provide an option to choose a custom path to get usage file. 

We will lose the flexibility to get help infos dynamically which can reflect code change. Eg: If we add a new option to `slc loopback:app`, we need to manually add it in `app.help.txt` to reflect the change then.

I was thinking of overriding function help in yoeman but then I realized as long as those infos are dynamically generated, they are not globalizable...

bajtos:
If this is the approach that apiconnect-loopback-cli is using, then I am ok with that for now.

In the future, if we ever get to refactor our generator to depend less on yeoman, I would prefer to introduce back an auto-generated help file reflecting code changes.

jannyHou:
>  but then I realized as long as those infos are dynamically generated, they are not globalizable...

Realized that It's a wrong assumption. I will override the help() function to generate help String.

jannyHou:
**Verify**

1. Install an old version of `generator-loopback@1.24.4`
2. Run `slc loopback:cmd_name --help` to print old help text. Copy them somewhere for comparison.

  A list of `cmd_name` to be tested:
  - [ ] app
  - [ ] acl
  - [ ] boot-script
  - [ ] datasource
  - [ ] export-api-def
  - [ ] middleware
  - [ ] model
  - [ ] property
  - [ ] relation
  - [ ] remote-method
  - [ ] swagger


3. Uninstall the old one and install `generator-loopback` with the latest commit.
4. Run same commands as step 2 to print new help text.
5. Compare the new version with old version, they should be same.

ssh24:
All the commands above have a missing space, other than that everything looks fine
Example with `swagger`
```
2c2
<   slc loopback:swagger [options] [<url>]
---
>  slc loopback:swagger [options] [<url>]
```

[...]

PR #229, opened 2016-09-23
Globalize/help files 

jannyHou:
connect to strongloop/generator-loopback#227
Changes in this pr:

- Create a new file `lib/help.js` to define our own function(keep the old name `customHelp`) to generate helpText. Now mostly functions are copied from [`yeoman: lib/actions/help.js`](https://github.com/yeoman/generator/blob/master/lib/actions/help.js) to keep it with the same file structure we have before:

  ```
    Usage:
    Options:
    Arguments:
    Other info from usage file
  ```

-  Move usage files into `intl/en` and name them as 'loopback_generatorName_usage.txt'

Now `customHelp()` will append text from 'intl/en/.txt' instead of file usage in each generator folder.


jannyHou:
@0candy Please take a look at the description above then review the pr. I separated changes into different commits, you can review them one by one for a easier understanding 😄 . 

The main changes are made in the first 3 commits. Others are fixes.

Thank you.

cc @bajtos

jannyHou:
Update: I just realized I didn't g.f() strings in new help.js.
Will do the globalization things if involved reviewers agree with this plan. I prefer to confirm this plan is reasonable first :).

lib/help.js
```diff
@@ -0,0 +1,141 @@
...
+  var helpStr = [
+    'Usage:',
+    ' ' + usageBuild(generator),
+    '',
+  ];
+
+  // Build options
+  if (Object.keys(generator._options).length) {
+    helpStr = helpStr.concat([
+      'Options:',
```
0candy: Needs to be globalized

lib/help.js
```diff
@@ -0,0 +1,141 @@
...
+      optionsBuild(generator),
+      '',
+    ]);
+  }
+
+  // Build arguments
+  if (generator._arguments.length) {
+    helpStr = helpStr.concat([
+      'Arguments:',
```
0candy: Needs to be globalized

lib/help.js
```diff
@@ -0,0 +1,141 @@
...
+function argumentsBuild(generator) {
+  var rows = generator._arguments.map(function(arg) {
+    var config = arg.config;
+
+    return [
+      '',
+      arg.name ? arg.name : '',
+      config.desc ? '# ' + config.desc : '',
+      config.type ? 'Type: ' + config.type.name : '',
+      'Required: ' + config.required,
```
0candy: Globalize `Required`

property/index.js
```diff
@@ -23,7 +24,7 @@ module.exports = yeoman.Base.extend({
   // saving project changes.
 
   help: function() {
-    return helpers.customHelp(this);
+    return helpText.customHelp(this, null, 'loobpack_property_usage.txt');
```
0candy: typo on loopback

jannyHou:
The test case is still in progress. I used `it.only` in the code to confirm that the rootDir is not setup correctly when invoking slc from test case. Still debugging how to set rootDir in test files.

jannyHou:
@0candy @bajtos I added test cases for all help commands. PTAL again. Thanks.


- Add test case

  >Have you considered adding tests for lib/help.js?

  Previously in `test/help.js` we test function `customHelp` by getting result from invoking command. So I do it in the same way to test all `--help` commands.

  I realize that I'd better add a function test for `customHelp` itself...will do it later.

- Bugs
  
   Command doesn't return all available generators in test case, I only put app generator in expected result to get test pass(otherwise other command will not be tested if app fails), see https://github.com/strongloop/generator-loopback/pull/229/commits/3e88730e6dff87f645c04ff53fb5035cd705de48#diff-0375403323a5c23fa036a172ed1ae25cR34 . I will dig into it then. 


- BCD License

   After I finished the open source course I believe we are able to override the function, but need to declare it. I prefer to spend time on it after the code review part done lol.

lib/help.js
```diff
@@ -0,0 +1,141 @@
...
+var table = require('text-table');
+var helpers = require('./helpers.js');
+
+/**
+ * Customize the help message based on how the generator is invoked (slc|yo)
+ * @param {Object} generator The generator instance
+ * @param {String} cmd The cmd name
+ * @returns {String} The customized help string
+ */
+exports.customHelp = function(generator, cmd, usageFilePath) {
```
bajtos: Can we change the signature to `function(generator, usageFilePath, cmd)`, so that callers can omit `cmd` instead of passing `null`?

```js
// ugly
return helpText.customHelp(this, null, 'loopback_datasource_usage.txt');

// better
return helpText.customHelp(this, 'loopback_datasource_usage.txt');
```

lib/helpers.js
```diff
@@ -141,35 +157,6 @@ exports.checkRelationName = function(modelDefinition, name) {
 };
 
 /**
- * Get the root command name based on the env var
- * @returns {string}
- */
-exports.getCommandName = function() {
```
bajtos: Is there any particular reason for moving the implementation of this function? Can we keep it in the old place, to reduce the patch size?

lib/help.js
```diff
@@ -0,0 +1,141 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: generator-loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+var g = require('strong-globalize')();
```
bajtos: I think you need to call `SG.SetRootDir`, don't you?

test/acl.test.js
```diff
@@ -6,6 +6,9 @@
 /*global describe, beforeEach, afterEach, it */
 'use strict';
 var path = require('path');
+var SG = require('strong-globalize');
+SG.SetRootDir(path.join(__dirname, '..'), {autonomousMsgLoading: 'all'});
```
bajtos: Please extract these two lines into `lib/globalize.js`, see how loopback does it in https://github.com/strongloop/loopback/blob/31913c048dfc88ba4993a6204d74bb45b5c32950/lib/globalize.js.

test/help.test.js
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
     process.env.SLC_COMMAND = undefined;
   });
 
+  it('print right help message for each generator', function() {
```
bajtos: `it('prints ...')`

test/help.test.js
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
...
+    process.env.SLC_COMMAND = 'loopback --help';
+    var names = ['acl', 'app', 'boot-script', 'datasource', 'export-api-def',
```
bajtos: Perhaps `command`, `cmd` or `cmdName` would be a more descriptive variable name?

test/help.test.js
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
...
+    'middleware', 'model', 'property', 'relation', 'remote-method', 'swagger'];
+
+    names.forEach(function(name) {
```
bajtos: Please create one test case for each name (command).

In your current implementation, the first failure will abort the whole test and remaining commands are not tested.

Here is a mock-up to show what I mean:

```js
['acl', ..., 'swagger'].forEach(function(cmdName) {
  it('prints correct help message for "yo ' + cmdName + '"', function() {
   var gen = givenGenerator(name, ['--help']);
   // ...
   assert.equal(output, helpText);
  }); 
});

File: test/help.test.js [hunk 12]
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
...
+      var gen = givenGenerator(name, ['--help']);
+      var helpText = gen.help();
+      var helpFileName = 'loopback_' + name + '_help.txt';
+      var helpFilePath = '../fixtures/' + helpFileName;
+      var output = fs.readFileSync(helpFilePath, 'utf8');
+      assert.equal(output, helpText);
+    });
+    process.env.SLC_COMMAND = undefined;
```
bajtos: This won't be run when the assertion above fails. Please move this line into a beforeEach or afterEach hook.

test/help.test.js
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
     process.env.SLC_COMMAND = undefined;
   });
 
+  it('print right help message for each generator', function() {
+    process.env.SLC_COMMAND = 'loopback --help';
```
bajtos: SLC_COMMAND is usually `slc` or `apic`, think about `$0` in shell (the name of the executable that started this process).

test/help.test.js
```diff
@@ -45,6 +46,22 @@ describe('loopback generator help', function() {
...
+    var names = ['acl', 'app', 'boot-script', 'datasource', 'export-api-def',
+    'middleware', 'model', 'property', 'relation', 'remote-method', 'swagger'];
+
+    names.forEach(function(name) {
+      var gen = givenGenerator(name, ['--help']);
+      var helpText = gen.help();
+      var helpFileName = 'loopback_' + name + '_help.txt';
+      var helpFilePath = '../fixtures/' + helpFileName;
+      var output = fs.readFileSync(helpFilePath, 'utf8');
+      assert.equal(output, helpText);
```
bajtos: The files in `intl` contain special directives for strong-globalize, e.g. `{{slc loopback}}`. Is it expected that the output of `gen.help()` includes them too and the return value of `gen.help()` has to be passed to strong-globalize?

As a user of this API, I would expect `get.help()` to return a text that was already processed/translated by strong-globalize.

Perhaps my understanding is wrong?

lib/helpers.js
jannyHou: Sure I will move it to the original line. I extracted it into new file then realized that old file still needs it, so moved it back, but seems not to the original line lol.

test/help.test.js
jannyHou: Now the return value of `gen.help()` is the **globalized** one, all strings in `lib/help.js` are globalized, e.g.: https://github.com/strongloop/generator-loopback/blob/globalize/help-files/lib/help.js#L48

So now we can compare `gen.help()` and the expected text directly.

jannyHou:
@bajtos I applied your feedback in the last 6 commits, hope the commit messages can reduce the review pain. PTAL again. Thanks!

cc @0candy

lib/help.js
```diff
@@ -0,0 +1,141 @@
...
+  if (generator._arguments.length) {
+    helpStr = helpStr.concat([
+      g.f('Arguments:'),
+      argumentsBuild(generator),
+      '',
+    ]);
+  }
+
+  // Append USAGE file if any
+  var exists = g.f(usageFilePath);
```
bajtos: I find it surprising that `g.f('loopback_remote-method_usage.txt')` returns `intl/en/loopback_remote-method_usage.txt`. Is this a feature of strong-globalize that `g.f` can "translate" filenames too?

@Setogit could you please confirm we are using strong-globalize correctly?

test/acl.test.js
bajtos: I think you should `require('../lib/globalize')` here too, unless I am missing something?

test/fixtures/loopback_acl_help.txt
```diff
@@ -0,0 +1,17 @@
+Usage:
+ slc loopback:acl [options]
```
bajtos: Can we move these files to a subdirectory, so that `test/fixtures` is not overcrowded? E.g. `test/fixtures/help-texts` or perhaps even `test/expected-help-texts`.

test/help.test.js
```diff
@@ -45,6 +46,26 @@ describe('loopback generator help', function() {
     process.env.SLC_COMMAND = undefined;
   });
 
+  describe('prints right help message for each generator', function() {
+    var cmdNames = ['acl', 'app', 'boot-script', 'datasource', 'export-api-def',
```
bajtos: Nitpick: we usually use all-upper-case style for constants.

```js
var CMD_NAMES = ...
```

test/help.test.js
```diff
@@ -45,6 +46,26 @@ describe('loopback generator help', function() {
...
+    'middleware', 'model', 'property', 'relation', 'remote-method', 'swagger'];
+
+    cmdNames.forEach(function(name) {
+      it('prints right help message for generator ' + name, function() {
+        process.env.SLC_COMMAND = 'slc loopback --help';
```
bajtos: See my older comment, I think this should use only `slc` as the command:

```js
process.env.SLC_COMMAND = 'slc';
```

Unless I am missing something?

test/help.test.js
```diff
@@ -45,6 +46,26 @@ describe('loopback generator help', function() {
...
+        var gen = givenGenerator(name, ['--help']);
+        var helpText = gen.help();
+        var helpFileName = 'loopback_' + name + '_help.txt';
+        var helpFilePath = '../fixtures/' + helpFileName;
+        var output = fs.readFileSync(helpFilePath, 'utf8');
+        assert.equal(output, helpText);
+      });
+    });
+    afterEach(function() {
```
bajtos: Nitpick: add an empty line before `afterEach` to serve as a visual delimiter. Also please use a named function, see http://loopback.io/doc/en/contrib/style-guide.html#layout-of-test-files

test/fixtures/loopback_acl_help.txt
jannyHou: Sure agree, I will create a subfolder for it.

And for
> I think you should require('../lib/globalize') here too, unless I am missing something?

I didn't require `lib/globalize` here since this file only needs `SetRootDir` but not `g.f()`(no need to import `g`), but on a second thought require glob file is easier and more consistent lol.

test/help.test.js
jannyHou: ah..I copied from https://github.com/strongloop/generator-loopback/blob/master/test/help.test.js#L32

Gave a quick look into `strongloop/lib` but still confused how the env works, will debug more about it.

`slc` works fine here. Will change.

lib/help.js
0candy: @bajtos strong-globalize returns the globalized help files as long as the file is in `intl/en`.
See example https://github.com/strongloop/strong-globalize#help-txt-files

jannyHou: IIRC https://github.com/strongloop/strong-globalize/blob/master/lib/globalize.js#L123 returns text if file found, while return file name as a pure string if file not found.

lib/globalize.js
```diff
@@ -0,0 +1,12 @@
+// Copyright IBM Corp. 2016. All Rights Reserved.
+// Node module: loopback
```
0candy: Should be `generator-loopback`

lib/help.js
bajtos: Cool, thank you for explaining. I find this part of strong-globalize API confusing, but that's out of scope of this patch.

jannyHou:
@0candy I fixed the module name. PTAL again. Thanks.

jannyHou:
@slnode test please

PR #232, opened 2016-10-11
Update pt translation file

0candy:


0candy:
@jannyHou PTAL

jannyHou:
Looked into node0.10 windows failure, but found nothing...no failing showed up even with "full log".
Not caused by this pr.

LGTM.

PR #239, opened 2016-10-22
Do not prompt for properties for KeyValueModel

superkhau:
Do not prompt for properties when running `slc loopback:model` and choosing `KeyValueModel` as the base model.

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1074

@bajtos PTAL

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -207,11 +207,18 @@ module.exports = yeoman.Base.extend({
   },
 
   delim: function() {
+    if (this.base === 'KeyValueModel')
+      return;
+
     this.log(g.f('Let\'s add some %s properties now.\n', this.displayName));
   },
 
   property: function() {
     var done = this.async();
+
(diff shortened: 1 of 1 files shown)


bajtos:
@slnode test please

superkhau:
>I have restarted build because CIS builder has failed previously. Pleas check CI outcome before landing.

TY for the restart, CI has been hanging as of late. It's all green now though, so we're good. :+1: 

>Also, should we modify property/index too, so that users cannot add properties via slc loopback:property to a KeyValue-based Model?

Good point, should do this too. Will submit follow up PR.

superkhau:
>Also, should we modify property/index too...

@bajtos What do you mean by index?

bajtos:
> > Also, should we modify property/index too...

> What do you mean by index?

Sorry, I meant `property/index.js`

Issue #231, opened 2016-10-06
`slc loopback:datasource` with `other` gives wrong instructions


superkhau:
When scaffolding new apps, `slc` asks:

```
? Enter the connector name without the loopback-connector- prefix:
```

It only work properly when you add the prefix not without (ie. you have to enter in `loopback-connector-redis` instead of just `redis`, same for the kv connectors).

I propose we change the message or fix the bug (either TOB or at least change the message for the time being if we don't find time to work on it soon).

@bajtos @0candy @jannyHou

bajtos:
+1, moved to `#tob` with `P1`

jannyHou:
I suggest rewording the msg to ask for **FULL** name. 

Some components also need full name, e.g.: loopback-component-storage

superkhau:
or module name/package.json/npmjs name instead of "full" which might be ambiguous.

jannyHou:
Closing since https://github.com/strongloop/generator-loopback/pull/242 checked in.

PR #242, opened 2016-11-09
Reword message for custom connector

jannyHou:
Connect to strongloop/generator-loopback#231

Details see https://github.com/strongloop/generator-loopback/issues/231#issuecomment-252158540


jannyHou:
@superkhau  PTAL. Thanks :)

cc @0candy

PR #247, opened 2016-12-05
Update Files to inject dynamic port

siddhipai:
Modified swagger/README.md
@superkhau PTAL


superkhau:
I don't think this is true for the README, loopback defaults to port 3000. I think you should be checking the tests for dynamic ports.

superkhau:
Closing as this change is not required.

PR #249, opened 2016-12-06
Update .travis.yml

bajtos:
- Drop Node v0.10 which is no longer covered by LTS.
- Add v0.12, v4 and v6


bajtos:
Closing in favour of #250

PR #255, opened 2016-12-15
Add support for SLC_COMMAND="loopback-cli"

bajtos:
In order to support the upcoming tool `loopback-cli`, we need to modify generator-loopback to recognize the new wrapper. This time it's more tricky, because we are not only replacing `yo` with e.g. `slc`, but also dropping `loopback:` prefix from generator names.

```
lb model -> yo loopback:model
```

Connect to strongloop-internal/scrum-loopback#1028

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -54,12 +54,20 @@ module.exports = yeoman.Base.extend({
 
   help: function() {
     var msgs = [helpText.customHelp(this, 'loopback_app_usage.txt')];
-    msgs.push(g.f('Available generators: \n\n  '));
-    msgs.push(Object.keys(this.options.env.getGeneratorsMeta())
+
+    var list = Object.keys(this.options.env.getGeneratorsMeta())
       .filter(function(name) {
         return name.indexOf('loopback:') !== -1;
-      }).join('\n  '));
-    return msgs.join('');
+      });
+    if (helpers.getCommandName() === 'loopback-cli') {
+      list = list.map(name => name.replace(/^loopback:/, 'lb '));
+      msgs.push(g.f('\nAvailable commands:\n\n'));
+    } else {
+      msgs.push(g.f('\nAvailable generators:\n\n'));
+    }
+
+    msgs.push(list.map(it => '  ' + it).join('\n'));
+    return msgs.join('') + '\n';
   },
 
   injectWorkspaceCopyRecursive: function() {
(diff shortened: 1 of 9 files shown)


lib/help.js
```diff
@@ -45,12 +45,19 @@ exports.customHelp = function(generator, usageFilePath, cmd) {
...
+  var usageText = g.f(usageFilePath);
+  if (usageText !== usageFilePath) {
+    helpStr.push(usageText);
   }
 
   helpStr = helpStr.join('\n');
+
+  if (command === 'loopback-cli') {
+    return helpStr
+      .replace(/ (yo|slc) loopback:/g, ' lb ')
```
ritch:
```
/ (yo|slc) loopback:/g, ' lb '
                   ^        ^                    
```
spacing seems off... worth double checking

lib/help.js
```diff
@@ -45,12 +45,19 @@ exports.customHelp = function(generator, usageFilePath, cmd) {
...
+      .replace(/ (yo|slc) loopback/g, ' lb');
```
ritch: no `:` here? not sure if that was intentional

lib/help.js
bajtos: This is intentional.

Before:

```
Lorem ipsum yo loopback:model [options]
           ^--replace--^
```

After

```
Lorem ipsum lb model [options]
           ^--^
```

lib/help.js
bajtos: Before

```
Lorem ipsum yo loopback:model [options]
Lorem ipsum yo loopback [options]
```

```
Lorem ipsum lb model [options]
Lorem ipsum lb [options]
```

This second line (L58) handles the case of the default generator with no `:`.

lib/help.js
ritch: 👍

ritch:
LGTM

lib/help.js
eddiemonge: `.replace(/ (yo|slc) loopback:?/g, ' lb');` will handle both cases

bajtos: > .replace(/ (yo|slc) loopback:?/g, ' lb'); will handle both cases

Notice that in the second case, we don't want space after `lb`. Sure, I can craft a more complex regexp, but I think it's not worth it. As a popular joke goes, "I wrote a regexp to solve my problem. Now I have two problems." 😉

app/index.js
```diff
@@ -253,7 +261,10 @@ module.exports = yeoman.Base.extend({
       this.log();
     } else {
       this.log(g.f('  Create a model in your app'));
-      this.log(chalk.green('    $ ' + cmd + ' loopback:model'));
+      if (cmd === 'loopback-cli')
```
superkhau: I think you're right! Too bad. ;( It's love/hate since we get some consistency. I don't like it when it doesn't let me `cb && cb()` either, but rules are rules. ;p

PR #263, opened 2017-01-19
[SEMVER-MAJOR] Expose enhanced Swagger to LoopBack code generation to CLI

raymondfeng:
### Description
See https://github.com/strongloop/loopback-swagger/pull/81

#### Related issues
https://github.com/strongloop/loopback-swagger/pull/81

- None

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "generator-loopback",
-  "version": "2.3.0",
+  "version": "3.0.0-alpha.1",
   "description": "Yeoman generator for LoopBack",
   "license": "MIT",
   "main": "app/index.js",
@@ -31,7 +31,7 @@
     "js-yaml": "^3.4.2",
     "lodash": "^4.16.1",
     "loopback-api-definition": "^3.0.0",
-    "loopback-swagger": "^3.0.0",
+    "loopback-swagger": "^4.0.0",
     "loopback-workspace": "^3.28.0",
     "mkdirp": "^0.5.1",
     "request": "^2.60.0",
--- a/swagger/index.js
+++ b/swagger/index.js
@@ -94,26 +94,6 @@ module.exports = yeoman.Base.extend({
       if (basePath.indexOf('/') === 0) {
         basePath = basePath.substring(1);
       }
-
-      var swaggerModel = 'swagger_' + basePath.replace(/\//g, '_');
-      self.modelNames.push(swaggerModel);
-      var modelDef = {
-        name: swaggerModel,
-        http: {
-          path: basePath,
-        },
-        base: 'Model',
-        facetName: 'server', // hard-coded for now
-        properties: {},
-      };
-      api.modelDefinition = modelDef;
-      self.modelDefs.push(modelDef);
-      self.modelConfigs.push({
-        name: swaggerModel,
-        facetName: 'server', // hard-coded for now
-        dataSource: null,
-        public: true,
-      });
     }
 
     // eslint-disable-next-line one-var
@@ -131,13 +111,17 @@ module.exports = yeoman.Base.extend({
           name: model.name,
           plural: model.plural,
           base: model.base || 'PersistedModel',
-          facetName: 'server', // hard-coded for now
+          facetName: 'common', // hard-coded for now
           properties: model.properties,
+        });
+        var tags = api.spec.tags || [];
+        var found = tags.some(function(t) {
+          return t.name === m;
         });
         self.modelConfigs.push({
           name: model.name,
           facetName: 'server', // hard-coded for now
-          public: true,
+          public: found,
         });
       }
     }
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "generator-loopback",
-  "version": "2.2.1",
```
bajtos: Please use `3.0.0-alpha.1`, to prevent our internal npm registry from containing a `3.0.0` version uploaded by our CI.

swagger/index.js
```diff
@@ -95,6 +95,7 @@ module.exports = yeoman.Base.extend({
         basePath = basePath.substring(1);
       }
 
+      /*
```
bajtos: Why is this block commented out? Either fix the problem or remove the unused code entirely.

swagger/index.js
```diff
@@ -131,13 +133,21 @@ module.exports = yeoman.Base.extend({
...
+          facetName: 'common', // hard-coded for now
           properties: model.properties,
         });
+        var found = false;
+        var tags = api.spec.tags || [];
+        for (var j = 0, k = tags.length; j < k; j++) {
+          if (tags[j].name === m) {
+            found = true;
+            break;
+          }
```
bajtos:
```js
const found = api.spec.tags.some(t => t.name === m);
```

swagger/index.js
```diff
@@ -131,13 +133,21 @@ module.exports = yeoman.Base.extend({
           name: model.name,
           plural: model.plural,
           base: model.base || 'PersistedModel',
-          facetName: 'server', // hard-coded for now
+          facetName: 'common', // hard-coded for now
```
bajtos: Why to change this?

If `server` is no longer a good default, then I think we should make this value configurable.

swagger/index.js
```diff
@@ -318,12 +328,24 @@ module.exports = yeoman.Base.extend({
...
+          if (self.selectedModels[m] !== SELECTED_FOR_UPDATE &&
+            self.selectedModels[m] !== SELECTED_FOR_CREATE) {
+            return done();
+          }
+          var modelDef = api.modelDefinition;
+          for (var i = 0, n = self.modelDefs.length; i < n; i++) {
+            if (m === self.modelDefs[i].name) {
+              modelDef = self.modelDefs[i];
+              break;
+            }
```
bajtos:
```js
const modelDef = self.modelDefs.find(d => d.name === m);
```

package.json
```diff
@@ -31,11 +31,12 @@
     "js-yaml": "^3.4.2",
     "lodash": "^4.16.1",
     "loopback-api-definition": "^3.0.0",
-    "loopback-swagger": "^3.0.0",
+    "loopback-swagger": "^4.0.0",
     "loopback-workspace": "^3.28.0",
     "mkdirp": "^0.5.1",
     "request": "^2.60.0",
     "strong-globalize": "^2.6.8",
+    "swagger-parser": "^3.4.1",
```
bajtos: AFAICT, this dependency is not used - please remove.

bajtos:
@raymondfeng please check CI results too, most of the builds failed AFAICT.

raymondfeng:
@bajtos Most issues are fixed. PTAL again.

swagger/index.js
bajtos: Oh, I see, you are defaulting to `api.modelDefinition` if no matching `self.modelDefs[i].name` is found.

How about the following then?

```js
const modelDef = self.modelDefs.find(d => d.name === m) ||
  api.modelDefinition;
```

My point here is that iterating arrays should be avoided in favour of `Array.prototype` functions - they make the code much easier to understand and prevent a whole range of errors.

swagger/index.js
bajtos: How about this comment? What is the rationale for changing the facet from `server` to `common`?

raymondfeng: Now the model definitions from Swagger are mapped to pure data models (sort of data transfer object) as the tagged operations are reflected into public models (controllers). It makes sense to use `common` instead of `server`.

raymondfeng:
@bajtos PTAL based on my last comment.

swagger/index.js
bajtos: To be honest, I am still not convinced, but don't want to block the progress here.

One last question - have you considered users upgrading their projects from the old generator to this new version? In my understanding, moving the generated code from `server` to `common` means that users will have to rewrite most of their operation implementation and move it from `server` to `common` too. OTOH, maybe that's not a concern because the layout of the generated code is already too different from the previous code-gen version?

bajtos: It would be great to provide a migration guide. We can review the instructions then to see if there are any opportunities to make the upgrade less cumbersome.

raymondfeng: The new swagger->loopback mapping will be released as a new major. It significantly changes the convention. We'll update the document to reflect that. There is not much migration we can do though.

bajtos:
LGTM, :shipit:

PR #264, opened 2017-01-20
Print IBM promotion even with --skip-next-steps

bajtos:
While fixing https://github.com/strongloop-internal/scrum-loopback/issues/1411, I noticed that the APIC promotion is not printed when the generator is run with `--skip-next-steps`.

In this patch, I am moving the APIC banner to a standalone method/async step, so that it is run even when next steps are skipped.

cc @smartmouse

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -238,48 +238,56 @@ module.exports = yeoman.Base.extend({
 
   installing: actions.installDeps,
 
-  end: function() {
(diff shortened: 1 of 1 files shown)


bajtos:
The changes are best viewed here: https://github.com/strongloop/generator-loopback/pull/264/files?w=1

bajtos:
Released in`generator-loopback@2.2.1`

PR #266, opened 2017-02-01
Update default LB version to 3.x

joesepi:
The generator still defaults to version 2.x rather than the current version 3.x


bajtos:
@joesepi good catch, thank you for the patch.

bajtos:
Released in `generator-loopback@2.3.0`

PR #277, opened 2017-04-12
Integrate datasource with lb soap feature

rashmihunt:
This PR has 
1)  support for  “lb soap” integration with SOAP WebServices datasource
2) support for updates if models already exist
3) Fixed Issue https://github.com/strongloop/loopback-cli/issues/28
4) Updated existing tests 
5) Updated README to point to loopback-cli README as single source of information. 
6) README in loopback-cli to change from WSDL URL - > datasource name will be done through another PR in loopback-cli repo
@raymondfeng PTAL

### DIFF
--- a/soap/index.js
+++ b/soap/index.js
@@ -45,17 +45,62 @@ module.exports = yeoman.Base.extend({
 
   loadProject: actions.loadProject,
 
-  askForWsdlUrlOrPath: function() {
-    var prompts = [
-      {
-        name: 'url',
-        message: g.f('Enter the WSDL url or file path:'),
-        default: this.url,
-        validate: validateUrlOrFile,
-      },
-    ];
+  loadDataSources: actions.loadDataSources,
+
+  addNullDataSourceItem: actions.addNullDataSourceItem,
+
+  loadModels: actions.loadModels,
+
+  existingModels: function() {
+    var self = this;
+    self.existingModels = this.modelNames;
+  },
+
+  checkForDatasource: function() {
+    var self = this;
+    var soapDataSourceNames = [];
+    var soapDataSources = [];
+    for (var i in this.dataSources) {
+      var datasource = this.dataSources[i];
+      if (datasource._connector === 'soap') {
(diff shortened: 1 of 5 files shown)


soap/index.js
```diff
@@ -45,17 +45,62 @@ module.exports = yeoman.Base.extend({
...
+    self.existingModels = this.modelNames;
+  },
+
+  checkForDatasource: function() {
+    var self = this;
+    var soapDataSourceNames = [];
+    var soapDataSources = [];
+    for (var i in this.dataSources) {
+      var datasource = this.dataSources[i];
+      if (datasource._connector === 'soap') {
```
raymondfeng: Do we need to check `loopback-connector-soap` too?

soap/index.js
```diff
@@ -45,17 +45,62 @@ module.exports = yeoman.Base.extend({
...
+  existingModels: function() {
+    var self = this;
+    self.existingModels = this.modelNames;
+  },
+
+  checkForDatasource: function() {
+    var self = this;
+    var soapDataSourceNames = [];
+    var soapDataSources = [];
+    for (var i in this.dataSources) {
```
raymondfeng: IIUC, `this.dataSources` is an array. Using `for...in` loop against an array is not practice. You should be able to use `this.dataSources.filter()` to get a subarray that matches `soap`.

soap/index.js
```diff
@@ -134,7 +181,7 @@ module.exports = yeoman.Base.extend({
 
     var api, i, n, m;
     self.operations = this.operations;
-    self.apis = generator.generateAPICode(this.operations);
+    self.apis = generator.generateAPICode(this.selectedDS.data.name, this.operations); // eslint-disable-line max-len
```
raymondfeng: Why the eslint override is needed? You can split the call in two lines as follows:
```
self.apis = generator.generateAPICode(this.selectedDS.data.name, 
  this.operations);
```

rashmihunt:
@raymondfeng Fixed review comments PTAL

PR #278, opened 2017-04-17
Print WSDL for the selected datasource

rashmihunt:
Usability improvement feedback from the demo which is to print the WSDL URL/file path once user selects the datasource. @raymondfeng PTAL

### DIFF
--- a/soap/index.js
+++ b/soap/index.js
@@ -101,6 +101,8 @@ module.exports = yeoman.Base.extend({
         }
       }
       self.url = self.selectedDS.data.wsdl;
+      self.log(chalk.green(g.f('WSDL for %s datasource: %s',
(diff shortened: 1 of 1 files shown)


soap/index.js
```diff
@@ -101,6 +101,8 @@ module.exports = yeoman.Base.extend({
         }
       }
       self.url = self.selectedDS.data.wsdl;
+      self.log(chalk.green(g.f('WSDL for %s datasource: %s',
```
raymondfeng: Maybe `g.f('WSDL for datasource %s: %s', ` is better

rashmihunt: @raymondfeng I had it like you suggested before, but after reading printed sentence, I changed it.. anyway, I am fine with either way.. changed it to your suggestion.

Issue #279, opened 2017-04-21
`yo loopback:model --bluemix` crashes when no Bluemix datasources are found


hacksparrow:
# Description/Steps to reproduce

`yo loopback:model --bluemix` crashes when no Bluemix datasources are found

1. Don't add any Bluemix datsource
2. Run `yo loopback:model --bluemix`

# Expected result

Should inform "Found no data sources to attach.."

hacksparrow:
#282 fixed this.

PR #282, opened 2017-04-25
Do not crash when Bluemix datasource are not found

hacksparrow:
#### Related issues

#279 

- None

### Checklist


hacksparrow:
@bajtos @raymondfeng guys, I need your input.

The model generator should not proceed ahead if no Bluemix datasource is defined. I can think of two ways of doing this.

- `return done('No Bluemix datasource found')` - async operation, cannot be asserted conventionally
- `process.exit(1)` - will exit `mocha` too; can't use this, if we want to add test cases

hacksparrow:
@bajtos thanks for the suggestion on Slack, I have made the changes accordingly to include the test case.

model/index.js
```diff
@@ -64,13 +60,41 @@ module.exports = yeoman.Base.extend({
 
   addNullDataSourceItem: actions.addNullDataSourceItem,
 
+  setBluemixDatasourceState: function() {
+    if (this.options.bluemix) {
+      var configPath = path.resolve(process.cwd(), '.bluemix',
```
raymondfeng: Let's use `this.destinationPath('.bluemix/datasources-config.json')` instead. See http://yeoman.io/authoring/file-system.html.

model/index.js
```diff
@@ -64,13 +60,41 @@ module.exports = yeoman.Base.extend({
...
+                      'datasources-config.json');
+      if (!fs.existsSync(configPath)) {
+        console.log('datasources-config.json not found');
```
raymondfeng: Use `this.log` instead.

model/index.js
```diff
@@ -64,13 +60,36 @@ module.exports = yeoman.Base.extend({
 
   addNullDataSourceItem: actions.addNullDataSourceItem,
 
+  setBluemixDatasourceState: function() {
+    if (this.options.bluemix) {
+      var configPath = this.destinationPath('.bluemix/datasources-config.json');
+      if (!fs.existsSync(configPath)) {
+        this.log('datasources-config.json not found');
+        process.exit();
```
raymondfeng: It's not ideal, but I cannot find a graceful way in Yeoman to report the error and terminate the generator. Unfortunately `done(err)` prints ugly stack trace. I'm not sure if `process.exit()` will mess up the terminal.

hacksparrow: Yes, I agree with the points.

No prompts are presented when `process.exit()` is called at that point, so it should be a clear exit. What other ways could this mess up the terminal?

Showing `done(err)`'s stack trace for that particular condition doesn't look like good UI. However, I decided to use `done(err)` on line no 85 to make the code testable.

PR #288, opened 2017-05-17
Add --login=false to avoid BM login for tests

raymondfeng:
### Description

When cached BM config is not found, the tests fail due to login requests without this change.

#### Related issues

- None

### Checklist


rashmihunt:
@slnode test please

PR #301, opened 2017-07-14
Handle special char in API filename

rashmihunt:
There are 2 bugs in this issue https://github.com/strongloop/loopback-connector-soap/issues/78 
and the first issue is fixed in this PR https://github.com/strongloop/loopback-soap/pull/7.  2nd issue is fixed with this current PR. We use binding name as top level REST API file/model filename and if special characters are present (e.g '.') which are not acceptable in filename, then, code generation fails with exception as reported in issue https://github.com/strongloop/loopback-connector-soap/issues/78 . This PR replaces special characters with underscore '_' . Also added test to cover this use case and cleaned up/refactored test code.
@raymondfeng PTAL

### DIFF
--- a/soap/index.js
+++ b/soap/index.js
@@ -190,8 +190,11 @@ module.exports = yeoman.Base.extend({
     for (i = 0, n = self.apis.length; i < n; i++) {
       api = self.apis[i];
       // TODO [rashmi] use binding name for now
-      var basePath = this.bindingName;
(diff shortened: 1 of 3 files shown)


rashmihunt:
@slnode test please

rashmihunt:
@hacksparrow Please kick off CI for this PR once you have bluemix issue fixed.

hacksparrow:
@rashmihunt merged.

PR #316, opened 2017-08-24
Fix export-api-def hanging if project contains ongoing async operations

Overdrivr:
### Description
If the command `export-api-def` is ran on an app that has ongoing async calls and never returns, the loopback generator itself never returns, waiting for module to finish loading.

This fix exits the Node process as soon as swagger spec has been written

WIP because for now I'm not sure how to test this fix. 

#### Related issues

- strongloop/loopback-cli/issues/50

### Checklist

### DIFF
--- a/export-api-def/index.js
+++ b/export-api-def/index.js
@@ -45,5 +45,8 @@ module.exports = yeoman.Base.extend({
     } else {
       process.stdout.write(apiDef);
     }
+
+    // Kill app if still alive
+    setTimeout(process.exit, 100);
   },
 });
--- a/test/export-api-def.test.js
+++ b/test/export-api-def.test.js
@@ -8,6 +8,7 @@ var path = require('path');
 var helpers = require('yeoman-test');
 var fs = require('fs');
 var expect = require('chai').expect;
+var sinon = require('sinon');
 var wsModels = require('loopback-workspace').models;
 var common = require('./common');
 var yaml = require('js-yaml');
@@ -16,6 +17,7 @@ var SANDBOX =  path.resolve(__dirname, 'sandbox');
 var PKG_CACHE = path.resolve(__dirname, '..', '.pkgcache');
 
 describe('loopback:export-api-def generator', function() {
+  var PROCESS_EXIT_STUB;
   this.timeout(30 * 60 * 1000); // 30 minutes
   before(common.resetWorkspace);
 
@@ -42,6 +44,14 @@ describe('loopback:export-api-def generator', function() {
 
   before(function installDependencies(done) {
     install(SANDBOX, PKG_CACHE, ['dependencies'], done);
+  });
+
+  beforeEach(function stubProcessExit() {
+    PROCESS_EXIT_STUB = sinon.stub(process, 'exit');
+  });
+
+  afterEach(function unstubProcessExit() {
+    PROCESS_EXIT_STUB.restore();
   });
 
   it('produces JSON format', function(done) {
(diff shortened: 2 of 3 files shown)


slnode:

slnode:
Can one of the admins verify this patch?

slnode:
Can one of the admins verify this patch?

export-api-def/index.js
```diff
@@ -45,5 +45,10 @@ module.exports = yeoman.Base.extend({
     } else {
       process.stdout.write(apiDef);
     }
+
+    // Kill app if still alive
+    if (!app.loaded) {
```
bajtos: I find this suspicious, why is this check needed? IMO, we should always exit.

export-api-def/index.js
```diff
@@ -45,5 +45,10 @@ module.exports = yeoman.Base.extend({
...
+      process.exit();
```
bajtos: I remember there were issues with writing to `stdout` and immediately exiting the process afterwards. When the output is longer, then on some platforms like Windows and some Node.js versions, the process will exit before the output was fully printed. See https://github.com/strongloop/loopback-sdk-angular-cli/commit/6a8c379bfc562e8715d3a63288cec29c74116940

While I am not sure if this is still an issue, perhaps we could add a small delay ~30ms before calling `process.exit`, to give the stdout stream a chance to write all data?

bajtos:
@Overdrivr thank you for the pull request, please take a look at my comments above.

If it becomes too difficult to write a good test for this change, then I am ok to land it even without test coverage.

bajtos:
@slnode ok to test

bajtos:
@Overdrivr We also need you to sign our CLA before we can accept your contribution, see https://cla.strongloop.com/agreements/strongloop/generator-loopback and http://loopback.io/doc/en/contrib/code-contrib.html#how-to-contribute-to-the-code

Overdrivr:
@bajtos Nice feedback as always (especially given that the fix is pretty much a one liner :)

For testing, since `process.exit` is stubed (no choice here otherwise the entire test suite is going down), I thought that there was no point in adding a blocking boot script. I've considered checking `process.exit` is called was enough. Is it ok for you ?

I'm removing the WIP status, let me know if you need additional changes or to rebase the branch.

Cheers

test/export-api-def.test.js
```diff
@@ -102,6 +104,25 @@ describe('loopback:export-api-def generator', function() {
...
   });
 
+  describe('running on project with onging async operations', function() {
+    var PROCESS_EXIT_STUB;
+
+    before(function stubProcessExit() {
+      PROCESS_EXIT_STUB = sinon.stub(process, 'exit');
+    });
+
+    it('does not hang', function() {
```
bajtos: Since this test is not actually inspecting an app that would hang, I find the current test name misleading. I am proposing to rename it to something like "it eventually exits the process".

test/export-api-def.test.js
```diff
@@ -102,6 +104,25 @@ describe('loopback:export-api-def generator', function() {
...
+      var gen = givenGenerator();
+      gen.run(function() {
+        expect(PROCESS_EXIT_STUB.called).to.be.equal(true);
```
bajtos: Interesting. How comes that the stub was already called, when there is a timeout of 30ms? Will it always be the case that the generator finishes only after the timeout was triggered, or is this introducing a possible false test failure when timing gets wrong? Maybe the test should periodically poll the value of `PROCESS_EXIT_STUB.called` and finish the test only after it changed the value to true? That would make the original test name "does not hang" sensible again.

test/export-api-def.test.js
```diff
@@ -102,6 +104,25 @@ describe('loopback:export-api-def generator', function() {
     });
   });
 
+  describe('running on project with onging async operations', function() {
+    var PROCESS_EXIT_STUB;
+
+    before(function stubProcessExit() {
```
bajtos: I think you need to stub `process.exit` for all tests (i.e. in the top-level `describe` block in this file). Otherwise the earlier tests exit the process before the test suit is over.

You can verify this fact by reviewing [CI build log](https://travis-ci.org/strongloop/generator-loopback/jobs/269885685) and searching for "ongoing" - the test output does not contain this word, which means the test was not run at all.

test/export-api-def.test.js
```diff
@@ -8,7 +8,9 @@ var path = require('path');
 var helpers = require('yeoman-test');
 var fs = require('fs');
 var expect = require('chai').expect;
+var sinon = require('sinon');
 var wsModels = require('loopback-workspace').models;
+var ws = require('loopback-workspace');
```
bajtos: AFAICT, `ws` is not used anywhere, can you please revert this line?

test/export-api-def.test.js
Overdrivr: I realized that for some reason, the test was always passing because I had forgotten the done callback. I'm really not sure why the test passed although the `process.exit` was clearly not called at the time of the assertion. 

I've updated the code with your approach by polling every 10 milliseconds, until the test timeouts or `process.exit` gets called. Also added a test that checks `process.exit` is not called right away.

I've also fixed your other comments, let me know if the changes suit you.

bajtos:
Sorry for the delay, I was on a sick leave and now my backlog grew out of control. I'll try to review your new version later this week.

Overdrivr:
Thanks for the rebase and feedback ;)

bajtos:
Landed, thank you for the contribution!

PR #320, opened 2017-09-27
Allow disableInclude and nestRemoting options

raymondfeng:
Prompt users to confirm `disableInclude` and `nestRemoting` options when a relation is defined.

### DIFF
--- a/relation/index.js
+++ b/relation/index.js
@@ -60,6 +60,21 @@ module.exports = yeoman.Base.extend({
       var msg = g.f('Model not found: %s', this.modelName);
       this.log(chalk.red(msg));
       this.async()(new Error(msg));
+    } else {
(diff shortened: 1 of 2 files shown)


relation/index.js
```diff
@@ -155,13 +171,39 @@ module.exports = yeoman.Base.extend({
           return answers.through && answers.throughModel === null;
         },
       },
+      {
+        name: 'customThroughModel',
+        message: g.f('Enter the model name:'),
+        required: true,
+        validate: validateRequiredName,
+        when: function(answers) {
```
jannyHou: Looks like a dup as the prompt above?

raymondfeng: Good catch. Fixed.

Issue #331, opened 2018-01-25
GVT: Hardcode issue in "apic lb -h"


dhmlau:
Original issue: https://github.ibm.com/velox/toolkit/issues/747
Hardcode issue in "apic lb -h"
please check below screen capture
![b17c2d42-fadd-11e7-954e-3aac687817a2](https://user-images.githubusercontent.com/25489897/35405972-ab81531e-01d5-11e8-9b32-3072e49fbd34.png)

Summary of the investigation with @raymondfeng:
- the highlighted messages are coming from Yeoman
- we can see if we can override those messages.

cc @raymondfeng @FrankDongNB

dhmlau:
What I got from @raymondfeng: 
- can possibly override `usage()` or `help()` in ttps://github.com/strongloop/generator-loopback/blob/master/lib/help.js.  
- use `skipMessage` option to skip yeoman message during npm install

dhmlau:
Fixed in https://github.com/strongloop/generator-loopback/pull/332

PR #332, opened 2018-01-27
GVT: Hardcode issue in "apic lb -h"

dhmlau:
Overwrite the default help messages from Yeoman for
- --help
- --skip-cache
- --skip-install

connected to: https://github.com/strongloop/generator-loopback/issues/331


dhmlau:
@bajtos , I found another way to update the option description based on the option names, in the function in here: https://github.com/strongloop/generator-loopback/blob/master/lib/help.js#L101-L128.  But thought that it's more hacky that way.  
If that's the preferred place to make the change, I'm good with that too.  Please advise.

bajtos:
>  I found another way to update the option description based on the option names, in the function in here: https://github.com/strongloop/generator-loopback/blob/master/lib/help.js#L101-L128. But thought that it's more hacky that way.

I think that may not work as intended. When an option is defined with a translated description, then `optionsBuild` will try to translate it again, won't it?

raymondfeng:
@dhmlau Instead of overriding `option()`, we can delete the built-in options before adding them back with translated description. For example,

```
delete this._options['help'];
delete this._options['skip-cache'];
delete this._options['skip-install'];
```

Or we can add a method:

```
overrideOption(name, value) {
  delete this._options[name];
  this.option(name, value);
}

this.overrideOption('help', ...);
```

dhmlau:
CI failed for `loopback-cli` because the message in the [output](https://github.com/strongloop/loopback-cli/blob/master/test/fixtures/help-lb.txt) has changed due to this PR.

dhmlau:
@raymondfeng @bajtos , thanks for your input.  I've made the change as suggested.  Please review again.  Thanks!

app/index.js
```diff
@@ -29,9 +29,28 @@ module.exports = yeoman.Base.extend({
       type: String,
     });
 
+    // delete the default options coming from Yeoman
+    delete this._options['help'];
```
raymondfeng: It will be nice to add a comment explaining why these options need to be deleted 1st

bajtos:
> CI failed for loopback-cli because the message in the output has changed due to this PR.

@dhmlau Please open a follow-up pull request to fix loopback-cli as soon as this patch is landed.

dhmlau:
@bajtos , you're right!!  Let me modify my PR.

PR #336, opened 2018-02-22
Add loopbackVersion and template options to skip prompts

raymondfeng:
Now we can use lb --loopbackVersion 3.x --template notes to scaffold an
app without prompting such questions.

#### Related issues

- connect to https://github.com/strongloop/loopback-cli/issues/61

### Checklist

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -46,6 +46,16 @@ module.exports = yeoman.Base.extend({
       type: Boolean,
     });
 
+    this.option('loopbackVersion', {
+      desc: g.f('Select the LoopBack version'),
+      type: String,
+    });
+
+    this.option('template', {
+      desc: g.f('Set up the LoopBack application template'),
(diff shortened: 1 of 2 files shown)


dhmlau:
2 questions: 
- Does that mean velox would need to call `lb --loopbackVersion 3.x`?
- What happen when 4.x is out, and we want to allow creation of LB3.x and LB4.x app but not LB2.x app?
Thanks.

raymondfeng:
> Does that mean velox would need to call lb --loopbackVersion 3.x?

Yes.

> What happen when 4.x is out, and we want to allow creation of LB3.x and LB4.x app but not LB2.x app?

LoopBack 4 uses a different CLI (`lb4` at the moment). We'll deal with that when it happens :-).

dhmlau:
Actually I forgot about the CI failure.

PR #350, opened 2018-06-16
Add --version to help options

crshnburn:
### Description

Add `--version` to the list of options when displaying help.

#### Related issues

- connect to strongloop/loopback-cli#35

### Checklist

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -60,6 +60,11 @@ module.exports = yeoman.Base.extend({
       desc: g.f('Set up as a Bluemix app'),
       type: Boolean,
     });
+
+    this.option('version', {
+      desc: g.f('Display version information'),
+      type: Boolean,
+    });
   },
 
   help: function() {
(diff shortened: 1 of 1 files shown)


bajtos:
Hello @crshnburn, thank you for the pull request!

The changes looks mostly good. However, I am concerned about implications for people using generator-loopback with yeoman.

```
$ lb --version
4.1.0 (generator-loopback@5.7.0 loopback-workspace@4.0.0)
$ yo loopback --version
1.8.5
```

Can we perhaps wrap the code adding `--help` option with an if statement that will run it only when generator-loopback is invoked from `lb` CLI? See https://github.com/strongloop/generator-loopback/blob/7f3fbab8a87227818275b10d697c339f234a5a52/app/index.js#L326-L328

Thoughts?

crshnburn:
Thanks for the review @bajtos I agree that the `--version` option should only be shown for users of the `lb` command so have put in the if statement as suggested.

bajtos:
Landed, thank you for the contribution! ❤️

PR #363, opened 2018-07-30
WIP: upgrade acl generator

jannyHou:
### Description

Not ready for review yet. Just get `acl` generator's 4 tests all pass, while haven't verified by running a cmd.

And the help function is broken at this moment. Working on the fix.

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] Code conforms with the [style


.eslintrc
```diff
@@ -6,5 +6,8 @@
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
     }]
+   },
+   "parserOptions": {
+      "ecmaVersion": 8
```
bajtos: We cannot use `async` functions in generator-loopback yet, we need to keep support for Node.js 6.x.

Please revert this change.

acl/index.js
```diff
@@ -12,49 +12,71 @@ var async = require('async');
...
-  },
+  // NOTE(janny) Working on fixing the help function
 
-  loadProject: actions.loadProject,
+  // help() {
+  //   return helpText.customHelp(this, 'loopback_acl_usage.txt');
+  // }
 
-  loadModels: actions.loadModels,
+  async ACLloadProject() {
```
bajtos: Ditto, we cannot use `async` functions in generator-loopback yet.

.eslintrc
jannyHou: That's great. I was going to update all the `this.async()` place to use `async await`. 
Reverting it.

app/index.js
```diff
@@ -270,94 +257,94 @@ module.exports = yeoman.Base.extend({
...
+      'http://ibm.biz/tryAPIC'
+    )));
+    this.log();
+  }
+};
 
 // Export it for strong-cli to use
 module.exports._package = pkg.name + ': ' + pkg.version;
 module.exports._yeoman = yeoman;
+module.exports._yeomanEnv = require('yeoman-environment');
```
jannyHou: This is related to `strongloop` and `loopback-cli`

app/index.js
```diff
@@ -13,15 +13,16 @@ var workspace = require('loopback-workspace');
 var Workspace = workspace.models.Workspace;
 var bluemix = require('../bluemix/helpers');
 var fs = require('fs');
-var actions = require('../lib/actions');
+// var actions = require('../lib/actions');
```
jannyHou: **In the first commit**, please DO NOT review the change in file `app/index.js`.

I have to update it to test the `acl` generator from cli, while the app generator is not finished.

.eslintrc
bajtos: I like the idea of getting rid of `this.async()` - if you like to do that then I fully support you in that change!

The trick is to use Promise APIs instead of async functions.

For example:

```ts
async function foo() {
  await bar();
  return 'result';
}
```

needs to be written in Promise-only styler as follows:

```ts
function foo() {
  return bar
    .then(() => 'result');
}
```

app/index.js
```diff
@@ -270,94 +257,94 @@ module.exports = yeoman.Base.extend({
...
+        'security options, please check out %s',
+      'http://ibm.biz/tryAPIC'
+    )));
+    this.log();
+  }
+};
 
 // Export it for strong-cli to use
 module.exports._package = pkg.name + ': ' + pkg.version;
 module.exports._yeoman = yeoman;
```
bajtos: Do we still need to export `yeoman`? Isn't `_yeomanEnv` all that loopback-cli needs?

jannyHou:
FYI `yeoman-generator` has an additional default option called `--force-install`, I feel it's ok to have it for each generator so I didn't remove it.

bluemix/index.js
```diff
@@ -76,20 +71,47 @@ module.exports = yeoman.Base.extend({
...
-  promptDefaultServices: bluemix.promptDefaultServices,
-  addDefaultServices: bluemix.addDefaultServices,
-});
+  };
+
+  loginToBluemix() {
+    this.login.apply(this);
+  };
+
+  BMpromptServiceName() {
```
dhmlau: nitpick: would it be better to have the function name as `BMPromptServiceName` instead of `BMpromptServiceName`? 
Same for the functions below.

datasource/index.js
```diff
@@ -19,54 +19,61 @@ var path = require('path');
...
 
     this.option('sso', {
-      desc: g.f('Log into Bluemix with SSO'),
+      description: g.f('Log into Bluemix with SSO'),
       type: Boolean,
     });
 
-    this.argument(g.f('name'), {
-      desc: g.f('Name of the datasource to create.'),
+    this.argument('name', {
```
dhmlau: Need to keep `g.f('name')` instead of `'name'`

test/bluemix.test.js
```diff
@@ -37,193 +37,222 @@ function itSkipIf(flag) {
...
-  it('should generate datasources.bluemix.js', function(done) {
-    var gen = givenBluemixGenerator('--force --bluemix');
-    helpers.mockPrompt(gen, {
-      enableManifest: false,
-    });
-    gen.run(function() {
-      ygAssert.file('./server/datasources.bluemix.js');
-      done();
-    });
+  it('should generate datasources.bluemix.js', function() {
```
dhmlau: we don't need to pass `done` in the parameter or call done() inside?

test/z-app.test.js
```diff
@@ -0,0 +1,391 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
```
dhmlau: the file name is called `z-app.test.js`.  Is it possible to rename the file name to fit what it's supposed to do?

app/index.js
```diff
@@ -11,20 +11,20 @@ var yeoman = require('yeoman-generator');
...
-module.exports = yeoman.Base.extend({
-  constructor: function() {
-    yeoman.Base.apply(this, arguments);
+module.exports = class AppGenerator extends ActionsMixin(yeoman) {
+  constructor(args, opts) {
+    super(args, opts);
 
-    this.argument(g.f('name'), {
-      desc: g.f('Name of the application to scaffold.'),
+    this.argument('name', {
```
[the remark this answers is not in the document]
jannyHou: Good catch.

acl/index.js
```diff
@@ -12,49 +12,69 @@ var async = require('async');
...
-
-  loadProject: actions.loadProject,
-
-  loadModels: actions.loadModels,
-
-  loadAccessTypeValues: function() {
+  }
+
+  loadProject() {
+    debug('loading project...');
```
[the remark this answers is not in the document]
jannyHou: @dhmlau I think the progress info like ^ is fine, since it doesn't contain any information.
I will double check all the places that prints variables....I don't see any sensitive info in the generator, while some probably prints long nested object. Will remove them.

app/index.js
```diff
@@ -270,94 +276,100 @@ module.exports = yeoman.Base.extend({
...
+  }
+
+  // end() {
+  //   var done = this.async();
+  //   this.end(done);
+  // }
+
+  bluemix() {
+    if (this.options.bluemix) {
+      this.log('\nBluemix configuration:');
```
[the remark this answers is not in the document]
jannyHou: no problem :)

bluemix/index.js
jannyHou: good catch.

test/bluemix.test.js
jannyHou: @dhmlau ah, don't need to. same as I put in comment https://github.com/strongloop/loopback-datasource-juggler/pull/1616#discussion_r209622203, mocha handles if a `Promise` returned.

test/z-app.test.js
jannyHou: @dhmlau run the `app.test.js` first causes other tests fail, I couldn't figure it out within one day so I prefer to leave it out of the scope of this PR :( I can create an issue to fix it afterwards.

dhmlau: i see. so the tests in this file is only a temp place to know what tests we're skipping now? i'm good with a follow up PR to handle the rest.

acl/index.js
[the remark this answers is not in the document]
dhmlau: ok.  I have this question because it may not seem useful in the future for debugging purpose.  If you think it's needed, i'm ok to keep it. thanks.

test/app.test.js
```diff
@@ -1,391 +0,0 @@
-// Copyright IBM Corp. 2014,2016. All Rights Reserved.
```
shimks: I think the workaround to make sure that this test is run last is poor. We should look to fix the bug that causes other tests to fail.

I remember having the same problem when I was investigating the test failures here. The hacky solution was to write `gen.emit('end')` after tests that involve error throwing. I remember you talking about a way to get the generator instance with the newest helper function, so I think this should be fixable.

test/bluemix.test.js
```diff
@@ -37,193 +37,222 @@ function itSkipIf(flag) {
...
+    return helpers.run(path.join(__dirname, '../bluemix'))
+      .cd(SANDBOX)
+      .withPrompts({
+        enableManifest: false,
+      })
+      .withOptions({
+        force: true,
```
shimks: why do we need these options?

test/export-api-def.test.js
```diff
@@ -55,83 +55,87 @@ describe('loopback:export-api-def generator', function() {
...
+    it('immediately does not exit the process', function() {
       expect(PROCESS_EXIT_STUB.called).to.equal(false);
-      gen.run(function() {
-        expect(PROCESS_EXIT_STUB.called).to.equal(false);
-        done();
-      });
+      return helpers.run(path.join(__dirname, '../export-api-def'))
+        .cd(SANDBOX)
+        .withOptions({
+          output: 'swagger.yaml',
```
shimks: I don't see this option in the previous code. Any reason we're introducing it here?

test/export-api-def.test.js
```diff
@@ -55,83 +55,87 @@ describe('loopback:export-api-def generator', function() {
...
-      var gen = givenGenerator();
       expect(PROCESS_EXIT_STUB.called).to.equal(false);
-      gen.run(function() {
-        // Periodically check process.exit was called until timeout
-        checkProcessExitCalled(done);
-      });
+      helpers.run(path.join(__dirname, '../export-api-def'))
+        .cd(SANDBOX)
+        .withOptions({
+          output: 'swagger.yaml',
```
shimks: same as comment above

test/fixtures/help-texts/loopback_app_help.txt
```diff
@@ -35,4 +36,4 @@ Example:
 
 Available generators:
 
-  loopback:app
+
```
shimks: This change looks skeptical

test/help.test.js
```diff
@@ -51,36 +51,40 @@ describe('loopback generator help', function() {
 
   it('prints help message with lb if invoked from loopback-cli', function() {
     process.env.SLC_COMMAND = 'loopback-cli';
-    var gen = givenGenerator('app', ['--help']);
+    var gen = givenGenerator('app');
     var helpText = gen.help();
     debug('--HELP TEXT--\n', helpText);
+    // Jannyhou: to be fixed!
```
shimks: Are these comments still relevant?

test/help.test.js
```diff
@@ -51,36 +51,40 @@ describe('loopback generator help', function() {
...
     assert(helpText.indexOf(' yo ') === -1,
       '"yo" should not be present');
     assert(helpText.indexOf('Available generators') === -1,
       '"Available generators" should not be present');
   });
 
+  // Jannyhou: to be fixed!
   describe('prints right help message for each generator', function() {
+    // var CMD_NAMES = ['acl'];
```
shimks: This comment is not needed

test/help.test.js
```diff
@@ -51,36 +51,40 @@ describe('loopback generator help', function() {
...
       it('prints right help message for generator ' + name, function() {
         process.env.SLC_COMMAND = 'slc';
-        var gen = givenGenerator(name, ['--help']);
+        var gen = givenGenerator(name);
         var helpText = gen.help();
         var helpFileName = 'loopback_' + name + '_help.txt';
         var helpFilePath = '../fixtures/help-texts/' + helpFileName;
         var output = fs.readFileSync(helpFilePath, 'utf8');
-        assert.equal(output, helpText);
+        assert.equal(helpText, output);
```
shimks: The arguments are switched

test/model.test.js
```diff
@@ -132,69 +122,81 @@ describe('loopback:model generator', function() {
...
+            plural: 'Reviews',
+          }).then(function() {
+            var modelConfig = readModelsJsonSync();
+            expect(modelConfig.Review.dataSource).to.eql('db1');
+          });
+      }
+    });
+
+    it('should set dataSource to db if it exists', function() {
+      return new Promise(function(resolve, reject) {
```
shimks: I think we can just return `wsModels.DataSourceDefinition.create`; I'm assuming the method has promise support

test/model.test.js
```diff
@@ -132,69 +122,81 @@ describe('loopback:model generator', function() {
...
+            plural: 'Reviews',
+          }).then(function() {
+            var modelConfig = readModelsJsonSync();
+            expect(modelConfig.Review.dataSource).to.eql('db');
+          });
+      }
     });
 
-    describe('with --bluemix', function() {
+    describe.skip('with --bluemix', function() {
```
shimks: From what I remember, this test might supposed to fail when run locally. Try getting rid of the skip and see if the test passes in the CI

test/relations.test.js
```diff
@@ -46,206 +46,185 @@ describe('loopback:relation generator', function() {
...
+          model: 'Part',
+          through: 'CarPart',
+          options: {
+            disableInclude: true,
+          },
+        });
       });
+  });
+
+  // requires generator-yeoman v0.17
```
shimks: Can we get rid of these comments now?

bluemix/helpers.js
```diff
@@ -10,218 +10,226 @@ var g = require('strong-globalize')();
...
-        appHostPrompt, appDiskQuotaPrompt);
-    }
-
-    var toolchainPrompt = {
-      name: 'enableToolchain',
-      message: g.f('Do you want to create toolchain files?'),
-      type: 'confirm',
-      default: true,
-      when: !this.bluemix,
+      var done = this.async();
```
virkt25: Not needed.

bluemix/index.js
```diff
@@ -60,13 +55,13 @@ module.exports = yeoman.Base.extend({
...
       default: false,
     });
-  },
+  };
 
-  help: function() {
+  help() {
     return helpText.customHelp(this, 'loopback_bluemix_usage.txt');
-  },
+  };
```
virkt25: don't need the `;`

boot-script/index.js
```diff
@@ -11,26 +11,27 @@ var helpText = require('../lib/help');
...
       type: String,
     });
-  },
+  };
 
-  help: function() {
+  help() {
     return helpText.customHelp(this, 'loopback_boot-script_usage.txt');
-  },
+  };
```
virkt25: You shouldn't need the `;`

boot-script/index.js
```diff
@@ -46,9 +47,9 @@ module.exports = yeoman.Base.extend({
       this.name = answer.name;
       done();
     }.bind(this));
-  },
+  };
```
virkt25: You shouldn't need the `;`

boot-script/index.js
```diff
@@ -62,15 +63,15 @@ module.exports = yeoman.Base.extend({
     return this.prompt(question).then(function(answer) {
       this.type = answer.type;
     }.bind(this));
-  },
+  };
```
virkt25: You shouldn't need the `;`

boot-script/index.js
```diff
@@ -62,15 +63,15 @@ module.exports = yeoman.Base.extend({
...
     var source = this.templatePath(this.type + '.js');
 
     // yeoman-generator 0.20.x doesn't like the leading /
     var targetPath = path.normalize('server/boot/' + this.name + '.js');
     var target = this.destinationPath(targetPath);
 
     this.copy(source, target);
-  },
-});
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -19,54 +19,61 @@ var path = require('path');
...
     });
 
     this.argument(g.f('name'), {
-      desc: g.f('Name of the datasource to create.'),
+      description: g.f('Name of the datasource to create.'),
       required: false,
       type: String,
     });
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -19,54 +19,61 @@ var path = require('path');
...
 
-  help: function() {
+  help() {
     return helpText.customHelp(this, 'loopback_datasource_usage.txt');
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -19,54 +19,61 @@ var path = require('path');
...
+  }
 
-  setAppName: function() {
+  setAppName() {
     // https://github.com/strongloop/generator-loopback/issues/38
     // yeoman-generator normalize the appname with ' '
     this.appName =
       path.basename(process.cwd()).replace(/[\/@\s\+%:\.]+?/g, '-');
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -95,9 +102,9 @@ module.exports = yeoman.Base.extend({
 
       done();
     }.bind(this));
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -111,9 +118,9 @@ module.exports = yeoman.Base.extend({
         this.name = props.name;
       }.bind(this));
     }
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -140,28 +147,30 @@ module.exports = yeoman.Base.extend({
         this.connector = props.customConnector || props.connector;
       }.bind(this));
     }
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -140,28 +147,30 @@ module.exports = yeoman.Base.extend({
...
 
-  loginToBluemix: function() {
-    if (this.options.bluemix) { bx.login.apply(this); }
-  },
+  loginToBluemix() {
+    if (this.options.bluemix) { this.login(); }
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -140,28 +147,30 @@ module.exports = yeoman.Base.extend({
...
-  generateBluemixFiles: bx.generateFiles,
+  generateBluemixFiles() {
+    this.generateFiles();
+  }
 
-  selectBluemixDatasource: function() {
+  selectBluemixDatasource() {
     if (this.options.bluemix) { ds.selectBluemixDatasource(this, g); }
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -140,28 +147,30 @@ module.exports = yeoman.Base.extend({
...
-  bindServiceToApp: function() {
+  bindServiceToApp() {
     if (this.abort) return;
     if (this.options.bluemix) {
       this.serviceBindingStatus = 'binding';
       ds.bindServiceToApp(this);
       this.serviceBindingStatus = 'bound';
     }
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -235,9 +244,9 @@ module.exports = yeoman.Base.extend({
       this.settings = props || {};
       reportWarnings();
     }.bind(this));
-  },
+  };
```
virkt25: You shouldn't need the `;`

datasource/index.js
```diff
@@ -278,14 +287,15 @@ module.exports = yeoman.Base.extend({
...
         var dependency = {};
         dependency[moduleVersion[0]] = moduleVersion[1];
+        debug('add dependency %s', util.inspect(dependency, {depth: null}));
         jsonfileUpdater(path.join(this.projectDir, 'package.json')).append(
           'dependencies', dependency, done
         );
       }
     }.bind(this));
-  },
+  };
```
virkt25: `;` isn't needed.

datasource/index.js
```diff
@@ -308,14 +318,14 @@ module.exports = yeoman.Base.extend({
         return done(err);
       }.bind(this));
     }
-  },
+  };
```
virkt25: `;` isn't needed.

datasource/index.js
```diff
@@ -308,14 +318,14 @@ module.exports = yeoman.Base.extend({
...
 
-  updatePipeline: function() {
+  updatePipeline() {
     if (this.abort) return;
     if (this.options.bluemix) { ds.updatePipeline(this); }
-  },
+  };
```
virkt25: `;` isn't needed.

datasource/index.js
```diff
@@ -324,7 +334,9 @@ module.exports = yeoman.Base.extend({
     } else {
       return;
     }
-  },
+  };
```
virkt25: `;` isn't needed.

datasource/index.js
```diff
@@ -324,7 +334,9 @@ module.exports = yeoman.Base.extend({
...
 
-  saveProject: actions.saveProject,
-});
+  saveProject() {
+    this.saveProjectForGenerator();
+  };
```
virkt25: `;` isn't needed.

export-api-def/index.js
```diff
@@ -26,13 +25,13 @@ module.exports = yeoman.Base.extend({
...
       defaults: '',
     });
-  },
+  };
 
-  help: function() {
+  help() {
     return helpText.customHelp(this, 'loopback_export-api-def_usage.txt');
-  },
+  };
```
virkt25: `;` isn't needed.

export-api-def/index.js
```diff
@@ -48,5 +47,5 @@ module.exports = yeoman.Base.extend({
 
     // Kill app if still alive
     setTimeout(process.exit, 100);
-  },
-});
+  };
```
virkt25: `;` isn't needed.

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
+module.exports = function ActionsMixin(baseClass) {
+  return class extends baseClass {
+    constructor(args, opts) {
+      super(args, opts);
+    }
+    // All actions defined in this file should be called with `this` pointing
+    // to a generator instance
 
-/**
+    /**
```
virkt25: extra space.

test/model.test.js
jannyHou: oh...bad news. I just realized I only fixed the composed bluemix with app and datasource, while there is one with model. Sigh.

bluemix/helpers.js
```diff
@@ -10,218 +10,226 @@ var g = require('strong-globalize')();
...
+            type: 'confirm',
+          },
+          {
+            name: 'enableAppMetrics',
+            message: g.f('Do you want to enable appmetrics?'),
+            type: 'confirm',
+          },
+        ];
+
+        var self = this;
```
virkt25: I don't think this is needed since we are binding to `this`.

bluemix/helpers.js
```diff
@@ -10,218 +10,226 @@ var g = require('strong-globalize')();
...
+        return this.prompt(prompts).then(function(answers) {
+          self.enableAutoScaling = answers.enableAutoScaling;
```
virkt25: AFAIK this can be `this.enableAutoScaling` instead as the function is bound to `this`.

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
+      var done = this.async();
+      return this.prompt([
+        {
+          name: 'dir',
+          message: g.f('Enter name of the directory to contain the project:'),
+          default: this.appname,
+        },
+      ]).then(function(answers) {
+        var dir = answers.dir;
+        self.dir = dir;
```
virkt25: Should be able to do `this.dir` as this function is bound to `this`.

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
 
-  return false;
-};
+      process.env.WORKSPACE_DIR = this.projectDir;
+
+      return false;
+    };
 
-/**
+    /**
```
virkt25: alignment

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
+            // default to 1st one
+            defaultDS = self.dataSources[0].value;
+          }
+        }
+        self.defaultDataSource = defaultDS;
+        done();
+      });
+    };
+
+    /**
```
virkt25: alignment

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
     }
-    self.defaultDataSource = defaultDS;
-    done();
-  });
-};
 
-/**
- * Modify the list of datasources created by {@link loadDataSources}
- * and append an item for `null` datasource.
+    /**
```
virkt25: alignment

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
+      }
+      var destPath = this.destinationPath(dest);
+      debug('template copy %s %s', this.templatePath(source), destPath);
+      this.fs.copyTpl(
+        this.templatePath(source),
+        destPath,
+        data || this,
+        options
+      );
+      // memfs.create().copyTpl(
```
virkt25: this can be removed.

app/index.js
```diff
@@ -65,11 +67,12 @@ module.exports = yeoman.Base.extend({
       this.option('version', {
         desc: g.f('Display version information'),
```
bajtos: I noticed the first argument, `name`, is using `description` property name; while all other options (not arguments!) are still using `desc` property name. Could you please check which form is the right one and ideally unify the implementation to use the same property name everywhere (if possible)?

app/index.js
```diff
@@ -270,94 +278,93 @@ module.exports = yeoman.Base.extend({
...
 
-        this.log(g.f('Next steps:'));
+  bluemix() {
+    if (this.options.bluemix) {
+      this.log(g.f('\nBluemix configuration:'));
+      this.composeWith(require.resolve('../bluemix'), this.options);
+    }
+  };
+  printNextSteps() {
+    if (!this.options.initBluemix) {
```
bajtos: Can we use early return here to reduce the level of indentation?

```js
if (this.options.initBluemix) return;
```

datasource/index.js
bajtos: I am surprised that `eslint` and/or `prettier` did not catch these extra `;`. Are we missing an eslint rule in our config?

test/app.test.js
bajtos: +1 for finding a more robust solution.

test/datasource.test.js
```diff
@@ -24,294 +24,279 @@ describe('loopback:datasource generator', function() {
...
-      pkg = JSON.parse(pkg);
-      // eslint-disable-next-line no-unused-expressions
-      expect(pkg.dependencies.lodash).to.exist;
-      done();
-    });
+  // Jannyhou: To be fixed (3)!
+  // I manually verified that by choosing "install connector",
+  // the custom module is intalled
+  // Here it fails since folder `node_modules` is not created
+  it.skip('should support custom connector', function() {
```
bajtos: Posting a line-level comment to remind us about a test that's skipped now.

```js
   // Jannyhou: To be fixed (3)!
   // I manually verified that by choosing "install connector",
   // the custom module is intalled
   // Here it fails since folder `node_modules` is not created
```

test/middleware.test.js
```diff
@@ -22,86 +22,79 @@ describe('loopback:middleware generator', function() {
...
-      params: '{"z": 1}',
+      return helpers.run(path.join(__dirname, '../middleware'))
+        .cd(SANDBOX)
+        .withPrompts({
+          name: 'my-middleware-3',
+          phase: 'routes',
+          paths: ['/x', '/y'],
+          params: '{"z": 1}',
+        }).then(function() {
+          // eslint-disable-next-line max-len
```
bajtos: Let's rewrite the next line so that it fits in our max line length.

```js
const middlewareConfig = readMiddlewaresJsonSync('server');
const newSources = Object.keys(middlewareConfig.routes);
```

test/model.test.js
bajtos: Assuming loopback-workspace is depending on LB 3.x internally (and not on LB 2.x), then yes, methods like `create` should support promises out of the box.

test/z-app.test.js
bajtos: I am not happy about moving renaming this test file from `app.test.js` to `z-app.test.js` and back, because it prevents `git` from rendering a useful `diff` showing the changes made alongside the transfer.

If it's too difficult to fix these test properly, then I am proposing the following workaround:
1. In this pull request, keep the file name as `app.test.js` and mark the tests as skipped via `describe.skip` at top-level.
2. In the next pull request, rename the file to `z-app.test.js` and remove `.skip`.

Or maybe do both changes in the same pull request, but split them into two commits.

Thoughts?

Also BTW, the test name `z-app.test.js` does not make this test file the last one, there is still `test/zosconnectee.test.js` that's run after this one.

jannyHou: @bajtos True the code compare is pretty bad :( skipping it makes sense to me 👌

lib/actions.js
```diff
@@ -7,187 +7,295 @@
...
+    /**
+ * Make some of the file API aware of our source/destination root paths.
+ * `copy`, `template` (only when could be applied/required by legacy code),
+ * `write` and alike consider.
+ *
+ * @param {String} source      Source file to copy from. Relative to this.sourceRoot()
+ * @param {String} destination Destination file to write to. Relative to this.destinationRoot()
+ */
+
+    copy(source, dest) {
```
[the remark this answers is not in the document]
jannyHou: @shimks good question. Let me elaborate this breaking change between `yeoman-generator@0.x` and `yeoman-generator@3.x`:

0.x has a function `directory` that copies the files from source to destination for you, see [code here](https://github.com/strongloop/generator-loopback/blob/master/app/index.js#L276). While 3.x doesn't have it and leaves users to do the file moving. 
That's why I added our implementation of `directory` in the `actions.js` file at the first place, then the other nested functions it calls.
`directory` --> `copy`  --> `template`

oracle/index.js
```diff
@@ -205,58 +195,56 @@ Please follow instructions at %s.', INSTALL_URL)));
...
       }
     }.bind(this));
-  },
+  };
 
   // Make sure checkDriver will be run after npm install
-  end: {
-    requireDriver: function() {
-      if (this.skip) return;
+  requireDriver() {
```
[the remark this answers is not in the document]
jannyHou: I think `end` is a phase that version 0.x allows multiple functions to be executed. So I just refactor it to be a function that gets called in the original sequence. And I don't see a difference from the old generator :-p

test/bluemix.test.js
jannyHou: The old generator is created by calling [`givenBluemixGenerator('--force --bluemix');`](https://github.com/strongloop/generator-loopback/blob/master/test/bluemix.test.js#L55), that's where the options set :)

test/fixtures/help-texts/loopback_app_help.txt
jannyHou: @shimks 
> loopback:app

That looks like a workaround to hack the old code. The available generators should be detected by [code](https://github.com/strongloop/generator-loopback/blob/78b61bcb0ebdde746350d7a02161a5fcfe070c0f/app/index.js#L75-L83), I don't remember why there was an extra `loopback:app` there. 
I think removing it is the expected behaviour.

test/help.test.js
jannyHou: removed.

test/help.test.js
jannyHou: @shimks Yep, the old code was wrong. Read the error message you will see, the first argument is expected result, while the second one is the actual result.

bluemix/helpers.js
jannyHou: @virkt25 It's get called on [Line 173](https://github.com/strongloop/generator-loopback/pull/363/files#diff-5ec15de4470e70e230e53c3a08375e18R173)

lib/actions.js
jannyHou: I am not sure why the linting format is so weird here, but eslint complains about the old code...

bluemix/helpers.js
jannyHou: yep 👍

app/index.js
jannyHou: @bajtos either of them works, see [yeoman-generator](https://github.com/yeoman/generator/blob/1cd1031eb1cc2ed0288265dd15ddc6cfbcffabb7/lib/index.js#L202)

And our [help message generator](https://github.com/strongloop/generator-loopback/pull/363/files#diff-2dc44d3980a09a85c3906ab18870bca8R109) 

But I agree that it's better to unify them. I will change `description` to `desc`.

test/datasource.test.js
jannyHou: Hmm @bajtos could you elaborate more about what does "line-level comment" mean?

jannyHou:
@virkt25 Thank you for pointing out those extra semicolons! I feel the eslint definitely has some problem, having the semicolon/removing it both pass `npm run lint`.

And according to [`eslint-config-loopback`](https://github.com/strongloop/eslint-config-loopback/blob/4b21c51c1c3610b98c02e3071460830556e6e95d/eslint.json#L58), *the expected behaviour is `["error", "always"]`, which requires the semicolon per statement.* What do you think?

If we decide to remove all extra semicolon, I can add rule `"semi": ["error", "always"]` in `.eslintrc` as a workaround if fixing the lint takes time.

virkt25:
I'm indifferent to the rule as long as it's applied consistently to the code base.

jannyHou:
ok I figured out why `eslint` didn't report error for those extra semicolons.
FYI 
`"no-extra-semi": "error"` checks the semicolon after each class method.(inside class)
`"semi": ["error", "always"]` only checks the semicolon after class definition.(outside class)

And removed all extra semicolons in this [commit](https://github.com/strongloop/generator-loopback/pull/363/commits/03bdc82e2b0bb8ad2bd47c7e51b94f7869d7cb5a), will keep it as a separate commit when merge PR.

datasource/index.js
jannyHou: @bajtos see comment https://github.com/strongloop/generator-loopback/pull/363#issuecomment-416374651

.eslintrc
```diff
@@ -5,6 +5,7 @@
       "ignoreComments": true,
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
-    }]
+    }],
+    "no-extra-semi": "error"
```
bajtos: Please move this configuration to https://github.com/strongloop/eslint-config-loopback so that it's eventually enabled for all LB (3.x) projects/repositories.

package.json
```diff
@@ -14,6 +14,7 @@
     "node": ">=6"
   },
   "scripts": {
+    "lint:fix": "eslint --ignore-path .gitignore --fix .",
```
bajtos: Have you considered re-using the cli args from `lint` script via `npm run lint --fix`?

test/model.test.js
```diff
@@ -196,65 +196,75 @@ describe('loopback:model generator', function() {
...
-          var modelGen = givenModelGenerator('--bluemix --login=false');
-          helpers.mockPrompt(modelGen, {
-            name: 'Product',
-          });
-
-          modelGen.run(function() {
-            expect(modelGen.abort).to.eql(true);
-            done();
-          });
+          var ctx = helpers.run(path.join(__dirname, '../model'));
```
bajtos: Use `const`, not `var` - see https://loopback.io/doc/en/contrib/style-guide-es6.html

test/model.test.js
```diff
@@ -196,65 +196,75 @@ describe('loopback:model generator', function() {
...
-        var modelGen = givenModelGenerator('--bluemix --login=false');
-        helpers.mockPrompt(modelGen, {
-          name: 'Product',
-        });
-
-        modelGen.run(function() {
-          expect(modelGen.abort).to.eql(true);
-          done();
-        });
+        var ctx = helpers.run(path.join(__dirname, '../model'));
```
bajtos: Use `const`, not `var` - see https://loopback.io/doc/en/contrib/style-guide-es6.html

test/model.test.js
```diff
@@ -196,65 +196,75 @@ describe('loopback:model generator', function() {
...
-        });
-
-        modelGen.run(function() {
-          assert('cloudant-demo-service' in modelGen.bluemixDataSourcesList);
-          var modelConfig = readModelsJsonSync();
-          expect(modelConfig.Product.dataSource).to
-            .eql('cloudant-demo-service');
-          done();
-        });
+        var ctx = helpers.run(path.join(__dirname, '../model'));
```
bajtos: Use `const`, not `var` - see https://loopback.io/doc/en/contrib/style-guide-es6.html

bluemix/helpers.js
```diff
@@ -10,218 +10,225 @@ var g = require('strong-globalize')();
...
+          name: 'enableDocker',
+          message: g.f('Do you want to create Dockerfile?'),
+          type: 'confirm',
+          default: true,
+          when: !this.bluemix,
+        };
+
+        prompts.push(toolchainPrompt, dockerPrompt);
+
+        this.prompt(prompts).then(function(answers) {
```
bajtos: A nitpick: can you rewrite this using an arrow function and remove `.bind(this)`?

```ts
return this.prompt(prompts).then(answers => {
  // ...
});
```

test/datasource.test.js
bajtos: Sorry for the confusion. By line-level I meant that the comment was attached to a particular source code line (`test/datasource.test.js:83`) and will be hidden by GitHub once you change that code; as opposed to a "regular" comment that stays in the top-level discussion & expanded forever.

Anyhow, my intention was to point out a comment you have added and which says that a skipped test needs to be fixed. IMO, something needs to be done about that comment before landing this pull request.

If you think fixing the skipped test is out of scope of this PR, then I am totally fine with that, but please reword the comment so that it does not look like something that was slipped through the review process.

For example:

```ts
// FIXME(jannyHou) This test fails because folder `node_modules` is not created.
// For now, verify this functionality manually by choosing "install connector"
// and checking that the custom module is installed.
```

lib/actions.js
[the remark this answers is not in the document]
b-admike: Should we write tests for these functions or are they already tested as part of the generator specific tests?

app/index.js
```diff
@@ -270,94 +278,92 @@ module.exports = yeoman.Base.extend({
       },
       done
     );
-  },
+  }
 
-  copyFiles: function() {
+  writing() {
```
b-admike: nitpick: can there be a better function name here?

app/index.js
```diff
@@ -270,94 +278,92 @@ module.exports = yeoman.Base.extend({
...
     this.directory('.', '.');
-  },
+  }
 
-  generateYoRc: function() {
+  // call `this.fs.commit` to finish moving the template files
+  // from memory to destination folder.
+  commit() {
+    var done = this.async();
+    this.fs.commit(done);
```
b-admike: From Yeoman docs:
>It is worth noting that although this.fs exposes commit, you should not call it in your generator. Yeoman calls this internally after the conflicts stage of the run loop.

http://yeoman.io/authoring/file-system.html. It's fine if we use it, but I was wondering whether there are alternatives.

lib/actions.js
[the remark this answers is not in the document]
jannyHou: @b-admike We have tests for helper functions but not actions. Testing actions require complicated setup for generator and sandbox.
And honestly I would prefer to not add such overhead effort in this PR :-p
Also, every generator requires file moving, so we have plenty acceptance tests to prove those actions work well :)

app/index.js
jannyHou: couldn't remember why I changed the name....reverted it.

app/index.js
jannyHou: @b-admike yep it's a workaround. I eventually calls it because there is no easy better solution for it :(

bluemix/index.js
```diff
@@ -76,20 +71,47 @@ module.exports = yeoman.Base.extend({
...
-  promptDefaultServices: bluemix.promptDefaultServices,
-  addDefaultServices: bluemix.addDefaultServices,
-});
+  }
+
+  loginToBluemix() {
+    this.login.apply(this);
+  }
+
+  BMPromptServiceName() {
```
raymondfeng: The method names should be camelCase, such as `bmPromptServiceName`.

raymondfeng:
@jannyHou Thank you for the bold efforts! Mostly LGTM.

lib/actions.js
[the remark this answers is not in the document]
b-admike: Sounds good to me!

test/app.test.js
```diff
@@ -14,8 +14,14 @@ var common = require('./common');
 var assert = require('assert');
 var expect = require('chai').expect;
 var fs = require('fs');
+var rimraf = require('rimraf');
+
+// Rename it to run as the last test
```
b-admike: nitpick: can we re-phrase this statement to start like this: `The test file is named app.test.js as a workaround for...`?

model/index.js
```diff
@@ -46,22 +46,36 @@ module.exports = yeoman.Base.extend({
     // A workaround to get rid of deprecation notice
     //   "generator#invoke() is deprecated. Use generator#composeWith()"
     // See https://github.com/strongloop/generator-loopback/issues/116
-    this.invoke = require('yeoman-generator/lib/actions/invoke');
-  },
+    // this.invoke = require('yeoman-generator/lib/actions/invoke');
```
virkt25: Can this comment be removed?

jannyHou: yep good catch, removed.

Issue #368, opened 2018-08-14
[GSSC-PreTVT]English string occurs when input lb acl in command line in JPN locale


gongkegongke:
Test Case: 11.020.060

![image](https://user-images.githubusercontent.com/34856053/44081194-ba2a626c-9fe0-11e8-91fd-6649d40cac00.png)
![image](https://user-images.githubusercontent.com/34856053/44081214-c66eec0a-9fe0-11e8-8fa9-87bb17d84a47.png)

Issue: English string occurs when input **lb acl** in command line

dhmlau:
PR #381

dhmlau:
The mentioned PR will fix the 2 issues:
- `(all existing models)` 
- `other`

dhmlau:
PR #381 is merged. Next is to wait for the translation file to come back from the Aug 21 drop.

dhmlau:
Translation for ja/zh/ko are in.  Released in `generator-loopback@5.9.1`

dhmlau:
All the translation files have been released.  Closing this ticket as resolved.

PR #381, opened 2018-08-17
Make select role prompt translatable for lb acl command

dhmlau:
### Description
Related to https://github.com/strongloop/generator-loopback/issues/368

When running `lb acl` command, further down the prompts it asked to select a role.  The `other` was not translated. 
44081214-c66eec0a-9fe0-11e8-8fa9-87bb17d84a47

Output after the fix:

Translation on the messages.properties is needed after this PR.


shimks:
The first image is in Japanese, but the second is in Chinese (or what I presume to be). Is this intentional?

dhmlau:
@shimks, you're right that the first one is Japanese (from the tester) and 2nd is Chinese (from me).  I set the locale to Chinese for testing so that I can understand. :)    The fix should be applicable to all languages ones the messages are translated.

dhmlau:
The `(all existing models)` should be translated as well.  Fixed it in the 2nd commit.

screen shot 2018-08-20 at 9 33 40 am

PR #395, opened 2018-11-08
honor app name in arguments

jannyHou:
### Description

Fixing the test failure in https://github.com/strongloop/loopback-cli/pull/71, it's also a missing feature.

#### Related issues

https://github.com/strongloop/loopback-cli/issues/76

### DIFF
--- a/model/index.js
+++ b/model/index.js
@@ -114,6 +114,16 @@ module.exports = class ModelGenerator extends ActionsMixin(yeoman) {
 
   askForName() {
     if (this.abort) return;
+
+    if (this.arguments && this.arguments.length >= 1) {
+      debug('model name is provided as %s', this.arguments[0]);
+      this.name = this.arguments[0];
+      var valid = validateRequiredName(this.name);
+      if (valid === true) return;
+      this.log(valid);
+      this.name = undefined;
+    }
+
     var prompts = [
       {
         name: 'name',
--- a/test/model.test.js
+++ b/test/model.test.js
@@ -112,6 +112,19 @@ describe('loopback:model generator', function() {
       });
   });
 
+  it('honors the model name in arg', function() {
+    return helpers.run(path.join(__dirname, '../model'))
+      .cd(SANDBOX)
+      .withArguments(['AdvancedProduct'])
+      .withPrompts({
+        dataSource: 'db',
+        base: 'Product',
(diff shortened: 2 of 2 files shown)


model/index.js
```diff
@@ -114,6 +114,18 @@ module.exports = class ModelGenerator extends ActionsMixin(yeoman) {
 
   askForName() {
     if (this.abort) return;
+
+    if (this.arguments && this.arguments.length >= 1) {
+      debug('model name is provided as %s', this.arguments[0]);
+      this.name = this.arguments[0];
+      var valid = validateRequiredName(this.name);
+      if (valid === true) return;
+      else {
```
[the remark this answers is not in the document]
bajtos: No need to use `else` after `return`.

```js
if (valid === true) return;
helpers.reportValidationError(valid, this.log);
this.name = undefined;
```

test/model.test.js
```diff
@@ -112,6 +112,19 @@ describe('loopback:model generator', function() {
...
+  it('honors the model name in arg', function() {
+    return helpers.run(path.join(__dirname, '../model'))
+      .cd(SANDBOX)
+      .withArguments(['Product'])
+      .withPrompts({
+        dataSource: 'db',
+        base: 'PersistedModel',
+      }).then(function() {
+        var product = readProductJsonSync();
+        expect(product).to.have.property('base', 'PersistedModel');
```
bajtos: I find this slightly confusing - how is the base property related to the model name provided via CLI arguments? Should you perhaps check that `name` property was set to `Product` instead?

bajtos: > The test before it sets the base to CustomModel, therefore my test reset it to PersistedModel and assert accordingly.

Could you please add a comment to explain this?

In general, tests should be independent on each other and don't rely on side effects.

> but even the test fail this assertion would still be true :)

Isn't the same true for your current test in case only this single test is executed and all other tests are skipped? Either via `it.only` or in CLI via `mocha -g "loopback:model generator.*honors the model name in arg"`.

I think you should explicitly set a different base model in this test, one that's not the default and also not used by any other tests around. Perhaps `Model` may work? Then verify that this explicit value was written to the file.

jannyHou: @bajtos Hmm I see your concern. I rewrote the test, hopefully it makes more sense.

PR #396, opened 2018-11-19
fix model generator

jannyHou:
### Description

Fix the broken model generator 

#### Related issues

https://github.com/strongloop/loopback-cli/issues/79

- [ ] ~New tests added or existing tests modified to cover all changes~ There is no good way to test the composed generator
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

#### lb model with multiple properties
```
jannyhous-mbp:testapp jannyhou$ lb model
? Enter the model name: lbmodel
? Select the datasource to attach lbmodel to: db (memory)
? Select model's base class PersistedModel
? Expose lbmodel via the REST API? Yes
? Custom plural form (used to build REST URL):
? Common model or server only? common
Let's add some lbmodel properties now.

Enter an empty property name when done.
? Property name: hello
? Property type: string
? Required? No
? Default value[leave blank for none]:

Let's add another lbmodel property.
Enter an empty property name when done.
? Property name: count
? Property type: number
? Required? No
? Default value[leave blank for none]:

Let's add another lbmodel property.
Enter an empty property name when done.
? Property name:
```

#### slc loopback:model with multiple properties

```
jannyhous-mbp:testapp jannyhou$ slc loopback:model
? Enter the model name: slcmodel
? Select the datasource to attach slcmodel to: db (memory)
? Select model's base class PersistedModel
? Expose slcmodel via the REST API? Yes
? Custom plural form (used to build REST URL):
? Common model or server only? common
Let's add some slcmodel properties now.

Enter an empty property name when done.
? Property name: hello
? Property type: boolean
? Required? No
? Default value[leave blank for none]:

Let's add another slcmodel property.
Enter an empty property name when done.
? Property name: hi
? Property type: string
? Required? No
? Default value[leave blank for none]:

Let's add another slcmodel property.
Enter an empty property name when done.
? Property name:
```

#### lb property

Make sure the property generator behaves the same

```
jannyhous-mbp:testapp jannyhou$ lb property
? Select the model: lbmodel
? Enter the property name: aprop
? Property type: string
? Required? No
? Default value[leave blank for none]:
```

#### slc looopback:property

Make sure the property generator behaves the same

```
jannyhous-mbp:testapp jannyhou$ slc loopback:property
? Select the model: slcmodel
? Enter the property name: aprop
? Property type: number
? Required? No
? Default value[leave blank for none]:
```

PR #397, opened 2018-11-20
remove .npmignore

jannyHou:
### Description

Remove the .npmignore file. see comment in https://github.com/strongloop/generator-loopback/pull/396#discussion_r234926169

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] Code conforms with the [style


bajtos:
@b-admike 

> actually do we want to remove `.npmignore`? I thought it is needed when publishing?

`slt-release`, the tool we use for publishing, is automatically creating `.npmignore` for us, taking `.gitignore` as the base and adding few more entries to ignore, for example `test` directory.

When we commit `.npmignore`, then `slt-release` will use the committed version and it's up to us to ensure `.npmignore` stays up-to-date when changing `.gitignore` file.

In LB4, we are using a different approach: instead of relying on `.npmignore` black-list of files to exclude, we provide a white-list of files to include via `package.json`'s property `files`.

PR #403, opened 2019-05-04
chore: update LTS status

dhmlau:
### Description
- Update `README` on LTS status. Related to https://github.com/strongloop/loopback/issues/4180
- add `"author": "IBM Corp.",` in package.json so that `slt copyright` command can generate the right header.
- Update copyrights year. Related to https://github.com/strongloop-internal/scrum-apex/issues/406

### DIFF
--- a/zosconnectee/index.js
+++ b/zosconnectee/index.js
@@ -1,3 +1,8 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
+// Node module: generator-loopback
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
 'use strict';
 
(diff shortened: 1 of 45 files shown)


zosconnectee/index.js
```diff
@@ -1,3 +1,8 @@
+// Copyright IBM Corp. 2017,2018. All Rights Reserved.
```
emonddr: @dhmlau  why not 2019? This goes for all instances of 2018.

dhmlau: AFAIK, the 2nd year in this copyright statement is the year where it's last updated.

emonddr: I see, ok. :)

PR #411, opened 2019-07-09
run api generator after app is booted

mrbatista:
#### Related issues

- connect to strongloop/loopback-boot#291

### Checklist

### DIFF
--- a/export-api-def/index.js
+++ b/export-api-def/index.js
@@ -35,17 +35,25 @@ module.exports = class ExportAPIDefGenerator extends ActionsMixin(yeoman) {
     var filePath = this.options.output;
     var options = {format: this.format, output: filePath};
     var app = require(this.destinationRoot());
-    var apiDef = apiGenerator.getApiDef(app, options);
 
-    // Print to console if no output file specified.
-    if (filePath) {
-      mkdirp.sync(path.dirname(filePath));
-      fs.writeFileSync(filePath, apiDef);
+    if (app.booting) {
+      app.on('booted', runGenerator);
     } else {
-      process.stdout.write(apiDef);
+      runGenerator();
     }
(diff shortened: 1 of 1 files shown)


export-api-def/index.js
```diff
@@ -35,17 +35,25 @@ module.exports = class ExportAPIDefGenerator extends ActionsMixin(yeoman) {
     var filePath = this.options.output;
     var options = {format: this.format, output: filePath};
     var app = require(this.destinationRoot());
-    var apiDef = apiGenerator.getApiDef(app, options);
 
-    // Print to console if no output file specified.
-    if (filePath) {
-      mkdirp.sync(path.dirname(filePath));
-      fs.writeFileSync(filePath, apiDef);
+    if (app.booting !== false) {
```
bajtos: I am not sure if this is the right check. Shouldn't we wait for `booted` event only when `app.booting` is set to `true`? I vaguely remember that older versions loopback-boot were not setting `app.booting` property at all when the boot process completed in the same tick.

```suggestion
    if (app.booting) {
```

What do you think?

See also:
- https://github.com/strongloop/loopback-sdk-angular-cli/blob/94bed58889f6e103415dc0dcf9f5175ff74a4539/bin/lb-ng.js#L45-L49
- https://github.com/strongloop/loopback-boot/pull/286

mrbatista: 👍

mrbatista:
@bajtos ping

bajtos:
LGTM 👍  The build failures look unrelated to your changes.

bajtos:
Landed, thank you for the contribution! ❤️ 

I have published the fix in `generator-loopback@6.1.1`, enjoy 🎉

